Issue #142, opened 2020-01-17
GridapTypes not printed correclty


fverdugo:
We still have the default verbose printing for `GridapType` objects

There is a `Base.` missing before `show` on line. This yould be the problem
https://github.com/gridap/Gridap.jl/blob/2a5681883d86a8a917b9d62f849db1631cfc1064/src/new/Helpers/Helpers.jl#L91

@victorsndvg 

Can you fix this and add a small test that checks that objects are printed as we expect?

victorsndvg:
Solved in commit 9fcceac. PR #145

fverdugo:
Closed via #145

PR #145, opened 2020-01-17
overwrite Base.show method for GridapType as suggested in #142

victorsndvg:
Solves #142

### DIFF
--- a/src/new/Helpers/Helpers.jl
+++ b/src/new/Helpers/Helpers.jl
@@ -88,7 +88,7 @@ tfill(v,::Val{3}) = (v,v,v)
 """
 abstract type GridapType end
 
-function show(io::IO,object::GridapType)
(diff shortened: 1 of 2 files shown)


fverdugo:
@victorsndvg,

In order to check if objects get displayed as we want, you can use these examples:

For `print()`:
https://github.com/gridap/Gridap.jl/blob/ab94dbbd1a0904395c969eaf473ee5ff5feadfe6/test/FESpacesTests/FESpacesTests.jl#L12

For `display()`: 
https://github.com/gridap/Gridap.jl/blob/ab94dbbd1a0904395c969eaf473ee5ff5feadfe6/test/FESpacesTests/FESpacesTests.jl#L14

victorsndvg:
done in 6aca8c8

Issue #210, opened 2020-03-12
Refactoring Gridap.TensorValues


fverdugo:
```julia

# types

abstract type MultiValue <: Number

struct VectorValue{D,T} <: MultiValue
  data::NTuple{D,T}
  function VectorValue{D,T}(data::NTuple{D,T}) where {D,T}
    new{D,T}(data)
  end
end

struct TensorValue{D1,D2,T,L} <: MultiValue
  data::NTuple{L,T}
  function TensorValue{D1,D2,T}(data::NTuple{L,T}) where {D1,D2,T,L}
    @assert L == D1*D2
    new{D1,D2,T,L}(data)
  end
end

struct SymTensorValue{D,T,L} <: MultiValue
  data::NTuple{L,T}
  function SymTensorValue{D,T}(data::NTuple{L,T}) where {D,T,L}
    @assert L == D*(D+1)/2
    new{D,T,L}(data)
  end
end

struct SymFourthOrderTensorValue{D,T,L} <: MultiValue
  data::NTuple{L,T}
  function SymFourthOrderTensorValue{D,T}(data::NTuple{L,T}) where {D,T,L}
    @assert L == (D*(D+1)/2)^2
    new{D,T,L}(data)
  end
end

```

```julia
# Construction

v = VectorValue( (1,2) )
v = VectorValue(1,2)

v = VectorValue{2}( (1,2) )
v = VectorValue{2}(1,2)

v = VectorValue{2,Int}( (1,2) )
v = VectorValue{2,Int}(1,2)

v = VectorValue{0,Int}( () )
v = VectorValue{0,Int}()

s = TensorValue(  (11,21,12,22) ) # Assumes square tensor
s = TensorValue(11,21,12,22) # Assumes square tensor

s = TensorValue{3,2}(  (11,21,31,12,22,32) )
s = TensorValue{3,2}(11,21,31,12,22,32)

s = TensorValue{3,2,Int}(  (11,21,31,12,22,32) )
s = TensorValue{3,2,Int}(11,21,31,12,22,32)

s = TensorValue{0,2,Int}(  () )
s = TensorValue{0,2,Int}()

s = SymTensorValue( (11,21,22) )
s = SymTensorValue(11,21,22)

s = SymTensorValue{2}( (11,21,22) )
s = SymTensorValue{2}(11,21,22)

s = SymTensorValue{2,Int}( (11,21,22) )
s = SymTensorValue{2,Int}(11,21,22)

[...]

PR #239, opened 2020-04-23
Refactor/tensor values

victorsndvg:
PR to solve #210


victorsndvg:
note comment at line 192 in `Operations.jl` file:

> #TODO: write det and inv function for small specific cases.

Issue #244, opened 2020-05-05
Change the type of sizes/partition member variables of CartesianDescriptor


amartinhuertas:
Change the types of the `sizes` and `partition` fields of `CartesianDescriptor` to tuples instead of points

Currently the (unique) inner constructor of `CartesianDescriptor`, available here:

https://github.com/gridap/Gridap.jl/blob/master/src/Geometry/CartesianGrids.jl#L22

clashes with

https://github.com/gridap/Gridap.jl/blob/master/src/Geometry/CartesianGrids.jl#L46

fverdugo:
Thanks for reporting!

fverdugo:
Closed via PR #246

PR #246, opened 2020-05-12
Change the type of sizes/partition member variables of CartesianDescriptor 

amartinhuertas:
See issue #244

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Gridap"
 uuid = "56d4f2e9-7ea1-5844-9cf6-b9c51ca7ce8e"
 authors = ["Santiago Badia <santiago.badia@monash.edu>", "Francesc Verdugo <fverdugo@cimne.upc.edu>"]
-version = "0.9.2"
+version = "0.10.0"
 
 [deps]
 BSON = "fbb218c0-5317-5bc6-957e-2ee96dd4b1f0"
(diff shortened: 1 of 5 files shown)


amartinhuertas:
Hi @fverdugo! You can proceed with your code review. One doubt that comes into my mind. The set of changes in this PR to which ChangeLog cathegory should go? My guess is that they should go to `Changed`, as we have modified the behaviour of a generic function which is in the public API of Gridap.jl, in particular, `CartesianDescriptor`. Agreed?

fverdugo:
> The set of changes in this PR to which ChangeLog cathegory should go? My guess is that they should go to Changed, as we have modified the behaviour of a generic function which is in the public API of Gridap.jl, in particular, CartesianDescriptor. Agreed?

Yes, moreover we are changing the types of the fields in the struct, which in this case are public fields.

fverdugo:
@amartinhuertas I have added type annotations to the arguments in the inner constructor of `CartesianDescriptor`. I have realized that these arguments cannot be of arbitrary type since some particular combinations can conflict with the method `CartesianDescriptor(pmin::Point,pmax::Point,partition)`

fverdugo:
@amartinhuertas Can you write a new entry in NEWS.md about this PR once it is merged?

amartinhuertas:
> @amartinhuertas Can you write a new entry in NEWS.md about this PR once it is merged?

Done in f560fa245e68284b06dfb8ebdb2d2899d26253fc

PR #267, opened 2020-05-27
Bug fixed `restrict(::AbstractArray,::TriangulationPortion)` for portions of BoundaryTriangulation(s)

amartinhuertas:
Hi @fverdugo,

when working with `GridapDistributed.jl`, I figured out that the (current) implementation of `restrict(::AbstractArray,::TriangulationPortion)`, available [here](https://github.com/gridap/Gridap.jl/blob/v0.10.2/src/Geometry/TriangulationPortions.jl#L47) is wrong whenever it is fed up with a portion of a triangulation that extends `BoundaryTriangulation` (or a portion of a portion of `BoundaryTriangulation`, a portion of a portion of a portion of `BoundaryTriangulation`, etc.). The current patch fixes the implementation of `restrict` in such cases. It also immediately fixes `restrict(::AbstractArray,::TriangulationPortion)` for portions of `SkeletonTriangulation`s, as these rely on portions of `BoundaryTriangulation's`. I tested that this fix works with `GridapDistributed.jl`, but I did not find a way to extend the current tests in `Gridap.jl` in order to  test this feature. Any idea?

@amartinhuertas

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
  - Fixed some methods of the `sparsecsr` generic function. Since PR [#262](https://github.com/gridap/Gridap.jl/pull/262).
+ - Fixed `restrict(::AbstractArray,::TriangulationPortion)` for portions of triangulations extending `BoundaryTriangulation`. Since PR [#267](https://github.com/gridap/Gridap.jl/pull/267).
 
 ## [0.10.2] - 2020-5-21
 
--- a/src/Geometry/BoundaryTriangulations.jl
+++ b/src/Geometry/BoundaryTriangulations.jl
@@ -67,6 +67,29 @@ function restrict(f::AbstractArray, trian::BoundaryTriangulation)
   compose_field_arrays(reindex(f,trian), get_face_to_cell_map(trian))
 end
 
+is_portion_of_boundary_triangulation(::TriangulationPortion{Dc,Dp,<:Triangulation}) where {Dc,Dp}=false
+is_portion_of_boundary_triangulation(trian::TriangulationPortion{Dc,Dp,<:TriangulationPortion}) where {Dc,Dp}=
+  is_portion_of_boundary_triangulation(trian.oldtrian)
+is_portion_of_boundary_triangulation(::TriangulationPortion{Dc,Dp,<:BoundaryTriangulation}) where {Dc,Dp}=true
+
+function get_face_to_cell_map(::TriangulationPortion{Dc,Dp,<:Triangulation}) where {Dc,Dp}
+  @assert false "You can only call get_face_to_cell_map with a TriangulationPortion of a BoundaryTriangulation"
+end
+function get_face_to_cell_map(trian::TriangulationPortion{Dc,Dp,<:TriangulationPortion}) where {Dc,Dp}
+  get_face_to_cell_map(trian.oldtrian)
+end
+function get_face_to_cell_map(trian::TriangulationPortion{Dc,Dp,<:BoundaryTriangulation}) where {Dc,Dp}
+  get_face_to_cell_map(trian.oldtrian)
+end
+
+function restrict(f::AbstractArray,trian::TriangulationPortion)
+  if (is_portion_of_boundary_triangulation(trian))
+    compose_field_arrays(reindex(f,trian), reindex(get_face_to_cell_map(trian),trian.cell_to_oldcell))
+  else
+    reindex(f,trian)
+  end
+end
+
 function get_cell_id(trian::BoundaryTriangulation)
   get_face_to_cell(trian)
 end
@@ -131,6 +154,3 @@ end
 #  D = num_cell_dims(model)
 #  face_to_mask = get_face_mask(labeling,tags,D-1)
 #end
-
-
-
--- a/src/Geometry/BoundaryTriangulations.jl
+++ b/src/Geometry/BoundaryTriangulations.jl
@@ -67,29 +67,6 @@ function restrict(f::AbstractArray, trian::BoundaryTriangulation)
   compose_field_arrays(reindex(f,trian), get_face_to_cell_map(trian))
 end
 
-# is_portion_of_boundary_triangulation(::TriangulationPortion{Dc,Dp,<:Triangulation}) where {Dc,Dp}=false
-# is_portion_of_boundary_triangulation(trian::TriangulationPortion{Dc,Dp,<:TriangulationPortion}) where {Dc,Dp}=
-#   is_portion_of_boundary_triangulation(trian.oldtrian)
-# is_portion_of_boundary_triangulation(::TriangulationPortion{Dc,Dp,<:BoundaryTriangulation}) where {Dc,Dp}=true
-#
-# function get_face_to_cell_map(::TriangulationPortion{Dc,Dp,<:Triangulation}) where {Dc,Dp}
-#   @assert false "You can only call get_face_to_cell_map with a TriangulationPortion of a BoundaryTriangulation"
-# end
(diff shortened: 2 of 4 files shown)


amartinhuertas:
> Can you check if it works?

I tested it. It did not work (see output below). Hard to say from this output why it did not work. You can try to reproduce the current error in your machine if u like. I pushed your proposed fix in the branch underlying this PR.

What's the rationale behind your proposed fix? (I do not full understand it)

```
include("test/DistributedPoissonDGTests.jl")
WARNING: replacing module DistributedPoissonDGTests.
Test Failed at 
/home/amartin/git-repos/GridapDistributed.jl/test/DistributedPoissonDGTests.jl:106
  Expression: e_l2 < tol
   Evaluated: 0.0002790769102222115 < 1.0e-9
ERROR: LoadError: There was an error during testing
in expression starting at 
/home/amartin/git-repos/GridapDistributed.jl/test/DistributedPoissonDGTests.jl:111
```

fverdugo:
Wow this is very strange... the tests in travis are passing.

In any case, I have fond a "little" error in the fix I proposed. 

@amartinhuertas can you try if the following works ?

```julia
function restrict(f::AbstractArray,trian::TriangulationPortion)
  reindex(restrict(f,trian.oldtrian),trian.cell_to_oldcell)
end
```
It also would be nice to find a test that only works with this last fix (if it ends up to be correct) and not with the previous one.

fverdugo:
> It also would be nice to find a test that only works with this last fix (if it ends up to be correct) and not with the previous one.

Try with a case in which `length(get_cell_id(trian.oldtrian))` is smaller than some id in `get_cell_id(trian)`

amartinhuertas:
>@amartinhuertas can you try if the following works ?

Great! Now it works. 

But I still do not understand the rationale behind your modified proposal. 

Can u please elaborate a little bit on it? 

Once I understand, then I can proceed with selecting a judicious test.

Thanks!

fverdugo:
> Can u please elaborate a little bit on it?

Sure!

It is important to note that we have 3 different sets of cells:

1. cells in the background interpolation mesh
2. cells in the  `oldtrian`
3. cells in the `trian`

Let's call `bgcell`, `oldcell`, and `cell` to a generic cell id for each of these sets of cells. Now, let me rewrite the function with this notation:

```julia
function restrict(bgcell_to_f::AbstractArray,trian::TriangulationPortion)
  oldcell_to_f = restrict(bgcell_to_f,trian.oldtrian)
  cell_to_f = reindex(oldcell_to_f,trian.cell_to_oldcell)
  return cell_to_f
end
```
Written in this way, the code is self reveling.

Another important thing to understand:

```julia
oldcell_to_bgcell = get_cell_id(oldtrian) 
cell_to_bgcell = get_cell_id(trian) 
```

amartinhuertas:
> Sure!
> It is important to note that we have 3 different sets of cells:
> ...

Thanks for the explanation! Now I got the point. Your modification is clearly the way to go. Let me complement your explanation with two extra points that I think are true and helped me to understand the whole picture (can you please explicitly confirm/check that they are true?):

1. Underlying the `restrict(f::AbstractArray,t::Triangulation)` generic function is the precondition that `f` has to be of the same length as the number of cells of dimension `Dc` in the DiscreteModel `t` stems from (either directly or indirectly). If you agree, perhaps we could add this statement as documentation for `restrict`, and/or as a precondition to be checked via `@assert`.

2. In the current (final) implementation of `restrict(f::AbstractArray,t::TriangulationPortion)`, the recursion which is triggered with `restrict(f,t.oldtrian)` ends up in two possible scenarios (are there more?):
   2.1.  `t.oldtrian` is the triangulation that directly stemmed from DiscreteModel, and of the same `Dc` as it. In such as case, `restrict(f,t.oldtrian)=f`.
   2.2.   `t.oldtrian` is a triangulation that extends `BoundaryTriangulation`. In such a case, `restrict(f,t.oldtrian)=compose_field_arrays(reindex(f,t.oldtrian),get_face_to_cell_map(t.oldtrian))`

@amartinhuertas

amartinhuertas:
> The tests are passing. However, I would wait to have the new test we were commenting so that we do not have a regression in this part.

Done in 829bdb0996bf3a531cf7a00daa8c738aada0bf96!

However, I have been trying to design a more clever test, without success. In particular, a test that spots the BUG that gives name to this PR: TriangulationPortion of BoundaryTriangulation. See the code I wrote so far below. The BUG was that the type of `v_btrian_portion.array` was wrong.  In particular, it ressembled more `v.array`, without being totally equal (because of a `ReindexedArray` in the middle of the awkarly large data type name). In other words, the test `@test typeof(v.array) != typeof(v_btrian_portion.array)` also passed with the BUG. How could we properly design the test?

```
domain = (0,1,0,1)
partition = (2,2)
model = CartesianDiscreteModel(domain,partition)
btrian = BoundaryTriangulation(model,[true for i=1:num_faces(model,1)])
btrian_portion = TriangulationPortion(btrian,[2,4,6])
btrian_portion_portion = TriangulationPortion(btrian_portion,[3,1])
order = 1
V0 = TestFESpace(
  reffe=:Lagrangian, order=order, valuetype=Float64,
  conformity=:H1, model=model, dirichlet_tags="boundary")
v=Gridap.FESpaces.get_cell_basis(V0)
v_btrian_portion=restrict(v,btrian_portion)
v_btrian_portion_portion=restrict(v,btrian_portion_portion)
@test length(v_btrian_portion)==3
@test length(v_btrian_portion_portion)==2
@test typeof(v.array) != typeof(v_btrian_portion.array)
```

fverdugo:
@amartinhuertas can you solve the conflict?

fverdugo:
> Underlying the restrict(f::AbstractArray,t::Triangulation) generic function is the precondition that f has to be of the same length as the number of cells of dimension Dc in the DiscreteModel t stems from (either directly or indirectly). If you agree, perhaps we could add this statement as documentation for restrict, and/or as a precondition to be checked via @assert.

Correct. Be also aware that Dc in the discrete model can be smaller than Dp in general.


> In the current (final) implementation of restrict(f::AbstractArray,t::TriangulationPortion), the recursion which is triggered with restrict(f,t.oldtrian) ends up in two possible scenarios (are there more?):
2.1. t.oldtrian is the triangulation that directly stemmed from DiscreteModel, and of the same Dc as it. In such as case, restrict(f,t.oldtrian)=f.
2.2. t.oldtrian is a triangulation that extends BoundaryTriangulation. In such a case, restrict(f,t.oldtrian)=compose_field_arrays(reindex(f,t.oldtrian),get_face_to_cell_map(t.oldtrian))

I think that in Gridap these are the two cases, however the number can increase since the code is extensible. E.g., in GridapEmbedded there are other cases.

amartinhuertas:
>@amartinhuertas can you solve the conflict?

Yes, done.

fverdugo:
> However, I have been trying to design a more clever test, without success. In particular, a test that spots the BUG that gives name to this PR: TriangulationPortion of BoundaryTriangulation. See the code I wrote so far below. The BUG was that the type of v_btrian_portion.array was wrong. In particular, it ressembled more v.array, without being totally equal (because of a ReindexedArray in the middle of the awkarly large data type name). In other words, the test @test typeof(v.array) != typeof(v_btrian_portion.array) also passed with the BUG. How could we properly design the test?

I think I will merge the PR even though this is not fixed since we need these devs in GridapDistributed ASAP. We can keep track of the pending test in an issue.

amartinhuertas:
> I think I will merge the PR even though this is not fixed since we need these devs in GridapDistributed ASAP. We can keep track of the pending test in an issue.

Ok. Agreed. Can u create the issue?

amartinhuertas:
> I think that in Gridap these are the two cases, however the number can increase since the code is extensible. E.g., in GridapEmbedded there are other cases.

May the BUG also affect to GridapEmbedded? Have you think about that?

PR #273, opened 2020-05-30
Vtk modification

oriolcg:
Hi @fverdugo , here is the pull request for the addition of the `createvtk` and `create_vtk_file` functions in `Vtk.jl`. The `writevtk` functions that contain loops inside have not been modified, assuming that for those cases the files are going to be always written within the loop.


oriolcg:
@fverdugo I added some tests that use the new functions within a `paraview_collection` workflow. The functions I added were already being stressed in the previous commit since they are called through `writevtk` and `write_vtk_file`, but now the testing is more complete.

Let me know if you think I should add something else.

PR #285, opened 2020-06-21
Bug in outer constructor of Table{...} for input arrays of abstract t…

amartinhuertas:
…ype.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [Unreleased]
+
+### Fixed
+
+  - Bug in outer constructor of Table{...} for input arrays of abstract type. Fixed via PR [#XXX](https://github.com/gridap/Gridap.jl/pull/XXX).
+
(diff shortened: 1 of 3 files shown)


amartinhuertas:
> Just a comment: be aware that this fix might allocate new vectors. 

Yes, sure!

> Perhaps in some contexts instead of using a Table, one could implement and use a more tailored specialization of a vector of vectors in order to avoid this.

Yes. Indeed, due to the fact that members of `Table` are of type `Vector`, then `Table` cannot serve our original purpose in `GridapDistributed.jl` (i.e., a vector of vectors build as a view of a one dimensional vector).  I had to apply an alternative solution not based in the use of `Table`. In any case, I applied boy scout's rule to fix this issue of `Table` that I found while reading the code bound to `Table`.

fverdugo:
:+1: 
If you have created some struct that is general enough, we can also include it in Gridap so that we can use it in other projects

amartinhuertas:
> If you have created some struct that is general enough, we can also include it in Gridap so that we can use it in other projects

I did not have to create an special structure. Three lines of code are sufficient:

https://github.com/gridap/GridapDistributed.jl/commit/e98d5f463cd3c6a9833373f4c3656758cac313f7#diff-fb2be2846b38d56cdaf599d624ece152R76

If u feel that this is going to be a recurrent/useful feature, then we can perhaps  develop a type extension of `AbstractVector`

Issue #297, opened 2020-06-26
Use promotion to build VectorValues and TensorValues from args with different types


fverdugo:
This is needed to fix

https://github.com/gridap/GridapODEs.jl/issues/14

fverdugo:
Closed via #298

PR #298, opened 2020-06-26
Use promotion in tensor values

fverdugo:


fverdugo:
Solves #297 and gridap/GridapODEs.jl#14

Issue #323, opened 2020-07-17
AssertionError triggered in the call path rooted at AppliedArrays: BUG?


amartinhuertas:
Hi @fverdugo, @santiagobadia,

while working in `GridapDistributed.jl`, I think I found a BUG in Gridap v.0.12.0 in the call path rooted at `function AppliedArray(g::AbstractArray,f::AbstractArray...)`. 

Let me call it BUG from now on, even if we don't still have a confirmation.

This BUG can be reproduced with the following minimal Julia code snippet, that synthetizes an actual call path status in a much more complex scenario (see below for a full Julia script in which this issue is spotted).

```julia
using Gridap
using FillArrays

g=FillArrays.Fill(Gridap.Fields.LinCom(),0)

#ptrs
j_to_i=Int64[]
i_to_v=[Int8(1) for i=1:24]
ptrs=Gridap.Arrays.reindex(i_to_v,j_to_i)
#values
values=[[0.7886751345948129 0.21132486540518708; 0.21132486540518713 0.7886751345948129]]
f1=Gridap.Arrays.CompressedArray(values,ptrs)

i_to_v=[[1, 2], [5, 6], [1, 5], [2, 6], [2, 3], [6, 7], [3, 7], [3, 4], [7, 8], [4, 8], [9, 10], [5, 9], [6, 10], [10, 11], [7, 11], [11, 12], [8, 12], [13, 14], [9, 13], [10, 14], [14, 15], [11, 15], [15, 16], [12, 16]]
j_to_i=Int64[]
lid_to_gid=Gridap.Arrays.reindex(i_to_v,j_to_i)
gid_to_val=Gridap.TensorValues.VectorValue{2,Float64}[(0.0, 0.25), (0.25, 0.25), (0.5, 0.25), (0.75, 0.25), (0.0, 0.5), (0.25, 0.5), (0.5, 0.5), (0.75, 0.5), (0.0, 0.75), (0.25, 0.75), (0.5, 0.75), (0.75, 0.75), (0.0, 1.0), (0.25, 1.0), (0.5, 1.0), (0.75, 1.0)]
f2=Gridap.Arrays.LocalToGlobalArray(lid_to_gid,gid_to_val)

Gridap.Arrays.AppliedArray(g,f1,f2)
```

The BUG comes in the form of an `AssertionError` triggered in a final node of the call path, namely, 

```julia
function _lincomb_checks(a,b)
  nb = length(b)
  np, na = size(a)
[...]

PR #329, opened 2020-07-22
Better handling of 0 length applied arrays

fverdugo:
Should fix issue #323


fverdugo:
@amartinhuertas I think I have fixed the issue. Can you check that the fix works for your case? If so, please merge the PR.

PR #354, opened 2020-08-11
add visualization_data

jw3126:


test/VisualizationTests/VisualizationDataTests.jl
```diff
@@ -0,0 +1,31 @@
...
+  conformity=:H1, model=model)
+U = TrialFESpace(V0)
+trian = Triangulation(model)
+degree = 2
+quad = CellQuadrature(trian,degree)
+A(u, v) = u ⊙ v
+b(v) = v⊙f
+t_Ω = AffineFETerm(A,b,trian,quad)
+op = AffineFEOperator(U,V0,t_Ω)
+u = solve(op)
```
jw3126, line 22: What is a simpler way to create fields then solving a FEM problem?

test/VisualizationTests/VisualizationDataTests.jl
```diff
@@ -0,0 +1,31 @@
...
+
+visdata = visualization_data(trian, cellfields=Dict("u" =>u))
+ncells = 12
+nodes_per_cell = 3
```
jw3126, line 26: How to extract these numbers directly from `trian` or `model`?

src/Visualization/VisualizationData.jl
```diff
@@ -0,0 +1,25 @@
+"""
+"""
+function visualization_data(trian::Triangulation; order=-1, nsubcells=-1, celldata=Dict(), cellfields=Dict())
```
jw3126, line 3: Should we mark this function experimental? I think we might want to change it in future. Like returning a `struct` instead of a named tuple etc.

test/VisualizationTests/VisualizationDataTests.jl
fverdugo: By interpolation
```julia
f(x) = sin(4*x[1])*cos(3*x[2])
u = interpolate(U,f)
```

In this case U can even be a test space without boundary conditions. You don't need to build a trial one.

jw3126: perfect thanks!

test/VisualizationTests/VisualizationDataTests.jl
fverdugo:
```julia
using Gridap.Geometry
using Gridap.ReferenceFEs
ncells = num_cells(vistrian)
nodes_per_cell = num_nodes(first(get_reffes(vistrian)))
```

src/Visualization/VisualizationData.jl
fverdugo: Don't worry about this.

test/VisualizationTests/VisualizationDataTests.jl
jw3126: Btw. it would be nice if the argument order of `interpolate` was switched. The usual convention is functions first, as it enables do block syntax.

fverdugo: > Btw. it would be nice if the argument order of interpolate was switched. The usual convention is functions first, as it enables do block syntax.

true.

jw3126:
I added `VisualizationData` struct. I think we will need to dispatch on something like `AbstractPlotting.plot!(::Plot(VisualizationData)` later.

jw3126:
@fverdugo can you merge this and add me to GridapMakie? I have some code I would like to put there.

fverdugo:
Hi @jw3126,

I would like to discuss what we want to put in GridapMakie before granting access. We can discuss it tomorrow e.g. via Gitter

PR #359, opened 2020-08-12
Remove unnecessary tfill methods

mohamed82008:
Only one edge case is needed I believe. I removed some unnecessary methods.


fverdugo:
These methods are just optimizations for some particular cases. They don't bother.

PR #360, opened 2020-08-12
Use tfill instead of fill in testvalue

mohamed82008:
This PR makes `testvalue` type stable.

### DIFF
--- a/src/Inference/Inference.jl
+++ b/src/Inference/Inference.jl
@@ -123,11 +123,11 @@ testvalue(::Type{T}) where T = zero(T)
 testvalue(v) = testvalue(typeof(v))
 
 function testvalue(::Type{T}) where T<:AbstractArray{E,N} where {E,N}
-   similar(T,tfill(0,N)...)
+   similar(T,tfill(0,Val(N))...)
 end
 
 function testvalue(::Type{T}) where T<:Fill{E,N} where {E,N}
-  Fill(zero(E),tfill(0,N)...)
+  Fill(zero(E),tfill(0,Val(N))...)
 end
 
 function testvalue(::Type{<:Tuple})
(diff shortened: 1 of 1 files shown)


src/Inference/Inference.jl
```diff
@@ -123,11 +123,11 @@ testvalue(::Type{T}) where T = zero(T)
 testvalue(v) = testvalue(typeof(v))
 
 function testvalue(::Type{T}) where T<:AbstractArray{E,N} where {E,N}
-   similar(T,fill(0,N)...)
+   similar(T,tfill(0,Val(N))...)
```
fverdugo, line 126: `Val(N)` is still type-unstable (unless the compiler is able to leverage that N is a compiler constant here).
Use `Val{N}()` instead.

src/Inference/Inference.jl
```diff
@@ -123,11 +123,11 @@ testvalue(::Type{T}) where T = zero(T)
...
 end
 
 function testvalue(::Type{T}) where T<:Fill{E,N} where {E,N}
-  Fill(zero(E),fill(0,N)...)
+  Fill(zero(E),tfill(0,Val(N))...)
```
fverdugo, line 130: Idem

src/Inference/Inference.jl
santiagobadia: It is the same

src/Inference/Inference.jl
santiagobadia: Idem

fverdugo:
Let's wait to some clarifying answer in 
https://discourse.julialang.org/t/should-i-use-val-n-or-val-n/44862

src/Inference/Inference.jl
mohamed82008: Since https://github.com/JuliaLang/julia/pull/24362 and Julia 1.0, we don't need to be so careful.

PR #365, opened 2020-08-13
add docstrings for residual and friends

jw3126:


src/FESpaces/FEOperators.jl
```diff
@@ -22,13 +22,26 @@ function allocate_residual(op::FEOperator,u)
   @abstractmethod
 end
 
+function doc_inplace_version(f)
+    # TODO move this helper elsewhere
+    "Inplace version of [`$f`](@ref)."
+end
```
jw3126, line 28: Suggestions where this helper function could live? Or is it over-engineered for it to exist?

santiagobadia: I feel it is a little bit over-engineering.
I would write it explicitly every time. It is about the same length.

src/FESpaces/FEOperators.jl
```diff
@@ -66,6 +91,12 @@ function residual_and_jacobian!(b::AbstractVector,A::AbstractMatrix,op::FEOperat
 end
 
 """
+    residual, jacobian = $(SIGNATURES)
+
+Compute the residual and jacobian of an operator `op` at a given point `u`.
+Depending on the nature of `op` the point `u` can either be a plain array or a `FeFunction`.
```
santiagobadia, line 97: FeFunction -> FEFunction

PR #366, opened 2020-08-13
switch argument order of interpolate to function first

jw3126:
This PR is should be against release 0.14 right? If that's correct, I will go ahead and fix the depwarns.


fverdugo:
Hi @jw3126 !

We can merge it against 0.14.

Since it is a new version, please replace the deprecation warnings by errors (with a meaning full error message) to make sure that the old notation is not used anymore from 0.14 on. 

Please add also a short note in NEWS.md in section `## Changed`

Thanks a lot!

jw3126:
> Since it is a new version, please replace the deprecation warnings by errors (with a meaning full error message) to make sure that the old notation is not used anymore from 0.14 on.

I think it would be better to do 0.14 depwarn and fully remove in 0.15. Note that you can do `julia --depwarn=error` to hunt down old notation.

fverdugo:
As you feel more confortable. I have no strong feelings on this. I just wanted to fix it out now.

In any case, what is important that all the test in the PR pass without dep. warnings.

jw3126:
Yeah, I use `julia --depwarn=error` locally to do this (we could add this to CI, but I think having it always active is more annoying than useful).

PR #376, opened 2020-08-24
Major refactoring associated with the skeleton and multi-field

fverdugo:


fverdugo:
cross-ref https://github.com/JuliaLang/julia/issues/37176

PR #392, opened 2020-08-27
document SkeletonCellField

jw3126:


jw3126:
BTW `SkeletonField` overloads `getproperty` such that `+, :inward, :left` are all aliases. I think ideally there is only one name for each property. And also if we really want to overload `getproperty` we should also overload `propertynames`.

fverdugo:
This PR is out-of-date after the last developments. Thanks for filing it anyway!

PR #396, opened 2020-09-05
Generalization of FESpaceWithLastDofRemoved

amartinhuertas:
As agreed, this PR:
 * Replaces `FESpaceWithLastDofRemoved` by `FESpaceWithDofPotentiallyRemoved`, a more general version with two additional features:
     1. One can select the dof to remove.
     2. One can activate or deactivate dof removal using a True/False trait (I know that you do not like True/False traits, but it in this case there are certainly only two options). 
* Modifies `ZeroMeanFESpace` such that it now relies on  `FESpaceWithDofPotentiallyRemoved`.

(not directly related to this PR, but FYI, `FESpaceWithDofPotentiallyRemoved` is already being succesfully used by the parallel distributed-memory version of `ZeroMeanFESpace`; click [here](https://github.com/gridap/GridapDistributed.jl/blob/e92b9c358df472ad0bcaa877cca7a7ebe66bdbfe/src/ZeroMeanDistributedFESpaces.jl#L99) for more details).

To simplify things, I think that two main concerns should be object to code review:
  1. A better name for `FESpaceWithDofPotentiallyRemoved`? (I used the one that we came up as a joke, but after thinking a bit, I have no better alternative).
  2. As the dof to remove may not be the first or the last dof, I had to introduce extra memory allocations wrt `FESpaceWithLastDofRemoved`. In particular, [here](https://github.com/gridap/Gridap.jl/blob/b3583187bf67a06913e73c6a2f41e816ac7b847a/src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl#L93) and [here](https://github.com/gridap/Gridap.jl/blob/b3583187bf67a06913e73c6a2f41e816ac7b847a/src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl#L109). Can we leave with that? If not, do you foresee any possible solution to avoid this?

@amartinhuertas


src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
```diff
@@ -0,0 +1,185 @@
+"""
+    struct FESpaceWithDofPotentiallyRemoved{CS,RD} <: SingleFieldFESpace
```
santiagobadia, line 2: `FESpaceConstantFixed` or  `FESpaceWithConstantFixed`

Note the difference between having the constants fixed and zero mean value

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
```diff
@@ -0,0 +1,185 @@
...
+  remove_dof::Val{RD}
+  dof_to_remove::Int
+  @doc """
+      FESpaceWithDofPotentiallyRemoved(space::SingleFieldFESpace, remove_dof::Bool,
+      dof_to_remove::Int=num_free_dofs(space))
+  """
+  function FESpaceWithDofPotentiallyRemoved(space::SingleFieldFESpace,
+     remove_dof::Bool,
+     dof_to_remove::Int=num_free_dofs(space))
+    s = "FESpaceWithDofPotentiallyRemoved can only be constructed from spaces without dirichlet dofs."
```
santiagobadia, line 21: Probably do nothing in this case. It reminds me Balancing Neumann-Neumann local spaces. If the constant is already fixed, do nothing. Minor point.

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
```diff
@@ -0,0 +1,185 @@
...
+  fv,dv,f::FESpaceWithDofPotentiallyRemoved{CS,true},cv) where {CS}
+  _fv, _dv = gather_free_and_dirichlet_values(f.space,cv)
+  @assert length(_dv) == 0
+  fv .= vcat(view(_fv,1:f.dof_to_remove-1),
+            view(_fv,f.dof_to_remove+1:length(_fv))) # TODO: can we avoid
+                                          # allocating new memory?
+  dv .= view(_fv,f.dof_to_remove:f.dof_to_remove)
+  (fv, dv)
+end
+
```
santiagobadia, line 115: In all these functions, we should decide which is the way to deal with the vector with one row eliminated in an efficient way. We could probably create a lazy vector doing this, using what we already have in Gridap. It is a quite obvious and would not involve any allocation instead of using views + vcat. Idem for the vector with only one touched value.

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
```diff
@@ -0,0 +1,185 @@
...
+      rj = -1
+    elseif bj < a.dof_to_remove
+      rj = bj
+    else
+      rj = bj-1
+    end
+    r[j] = rj
+  end
+  r
+end
```
santiagobadia, line 185: Do we need this? Couldn't we just change the `data` vector in the `Table` that represents the face dofs using a lazy vector that only touches one entry?

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
```diff
@@ -0,0 +1,185 @@
...
+      constraint_style::Val{CS}
+      remove_dof::Val{RD}
+      dof_to_remove::Int
+    end
+"""
+struct FESpaceWithDofPotentiallyRemoved{CS,RD} <: SingleFieldFESpace
+  space::SingleFieldFESpace
+  constraint_style::Val{CS}
+  remove_dof::Val{RD}
+  dof_to_remove::Int
```
santiagobadia, line 13: I am not sure you need `constraint_style` and `remove_dof`, already in the param.

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
amartinhuertas: I will use `FESpaceWithConstantFixed`. The potentially sense of the type is lost, but we can leave with that.

> Note the difference between having the constants fixed and zero mean value

Ok, good point. This space lets you fix the constant to an arbitrary value. The zero mean space fixes it to a particular value, i..e, the one which results in a zero mean function.

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
amartinhuertas: Proposal (feedback welcomed):

```
abstract type ConstantStrategy end;
type FixConstant           <: ConstantStrategy
type DoNotFixConstant <: ConstantStrategy
```

With `constraint_style` I need help. May be @fverdugo can help here? What does `True/False` mean for the `constraint_style` of FE spaces? On the other hand, I guess this is a cross-cutting change, right?

fverdugo: > With constraint_style I need help. May be @fverdugo can help here? What does True/False mean for the constraint_style of FE spaces? On the other hand, I guess this is a cross-cutting change, right?

`constraint_style(V) == Val(false)` for un-constrained spaces

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
```diff
@@ -0,0 +1,185 @@
...
+
+get_dirichlet_dof_tag(f::FESpaceWithDofPotentiallyRemoved{CS,false}) where {CS} = Int8[]
+
+function scatter_free_and_dirichlet_values(
+  f::FESpaceWithDofPotentiallyRemoved{CS,true},fv,dv) where {CS}
+  @assert length(dv) == 1
+  _dv = similar(dv,eltype(dv),0)
+  _fv = vcat(view(fv,1:f.dof_to_remove-1),
+             dv,
+             view(fv,f.dof_to_remove:length(fv))) # TODO lazy append
```
fverdugo, line 79: Try `Gridap.Arrays.lazy_append`

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
fverdugo: I think we need this (or something in this direction) to be general. We cannot assume in general that we have a `Table` for the face dofs. We could write (on top of this) a specialized version when we have a Table, but it seems premature optimization to me.

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
```diff
@@ -0,0 +1,185 @@
...
+end
+
+function gather_free_and_dirichlet_values(
+  f::FESpaceWithDofPotentiallyRemoved{CS,false},cv) where {CS}
+  gather_free_and_dirichlet_values(f.space,cv)
+end
+
+function gather_free_and_dirichlet_values!(
+  fv,dv,f::FESpaceWithDofPotentiallyRemoved{CS,true},cv) where {CS}
+  _fv, _dv = gather_free_and_dirichlet_values(f.space,cv)
```
fverdugo, line 107: Can we reduce code duplication between `gather_free_and_dirichlet_values` and `gather_free_and_dirichlet_values!` ?

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
fverdugo: The other option is to implement a `VectorWithIndexInserted`, or whatever name you like.

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
fverdugo: Another option is to implement a `VectorWithIndexRemoved` in addition to the `VectorWithIndexInserted` I commented above.

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
fverdugo: > This space lets you fix the constant to an arbitrary value.

FYI, the zero mean FE space currently assumes that the value is fixed exactly to 0 in the fixed dof of the `FESpaceWithDofPotentiallyRemoved`.

amartinhuertas: >FYI, the zero mean FE space currently assumes that the value is fixed exactly to 0 in the fixed dof of the FESpaceWithDofPotentiallyRemoved.

I do not see this statement. Doesn't it depend on the dirichlet values that come along with the trial FE space?

amartinhuertas: Done in faa32a397b4fdb261612c37fad0d73a2e5ce2658

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
amartinhuertas: Done (partially) in 0da971d918e87711098e3a32e14036c9741a4cde. In particular, I would prefer to transform `constraint_style` into struct-like traits in a different PR. I expect that such implies a number of changes that deserve its separate PR.

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
santiagobadia: It does not have to be a table but just create a method that modifies the `face_own_dofs` whatever it is. The idea is simple, only modify the plain global vector by touching one of, in the line of what you say above, e.g., `VectorWithIndexChanged` (horrible name) with the global Ids that is being used in `face_own_dofs`. This way, we do not need the code above, I guess.

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
amartinhuertas: Done in 0da971d.

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
amartinhuertas: I will implement `VectorWithIndexRemoved` and `VectorWithIndexInserted` because it will be even more efficient than using lazy vectors, and can be useful for other scenarios as well. Besides, lazy vectors do not currently support write operations, broadcasted assignments, etc. Is there any fundamental reason behind this? or it is just because it is missing? I do not need it, anyway.

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
amartinhuertas: > Can we reduce code duplication between gather_free_and_dirichlet_values and gather_free_and_dirichlet_values! ?

I guess the comment has become outdated after latest changes. Agreed? (See below)

```julia
function gather_free_and_dirichlet_values(
  f::FESpaceWithConstantFixed{CS,FixConstant},cv) where {CS}
  _fv, _dv = gather_free_and_dirichlet_values(f.space,cv)
  @assert length(_dv) == 0
  fv = VectorWithEntryRemoved(_fv,f.dof_to_fix)
  dv = _fv[f.dof_to_fix:f.dof_to_fix]
  (fv, dv)
end

function gather_free_and_dirichlet_values!(
  fv,dv,f::FESpaceWithConstantFixed{CS,FixConstant},cv) where {CS}
  _fv, _dv = gather_free_and_dirichlet_values(f.space,cv)
  @assert length(_dv) == 0
  fv    .= VectorWithEntryRemoved(_fv,f.dof_to_fix)
  dv[1]  = _fv[f.dof_to_fix]
  (fv, dv)
end
```

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
amartinhuertas: Done in PR https://github.com/gridap/Gridap.jl/pull/401

fverdugo (approved):

src/FESpaces/FESpacesWithDofPotentiallyRemoved.jl
amartinhuertas: I worked on all the issues, except this one.

Just to be double-check, you propose to remove the `CellDofsWithDofFixed` data type, right?

Plus a possible alternative solution, that I do not enterily grasp. What I do not get is:
  *  "Probably create a method for adding these constraints in face_own_dofs Table."
  *  "It does not have to be a table but just create a method that modifies the face_own_dofs whatever it is." 

`face_own_dofs` is a Table returned by `ReferenceFE` right? How is the link among this Table and `get_cell_dofs()` such that we can achieve the desired effect?.  I am sure that they key is here, and thus I am missing something important.

I have been exploring `get_cell_dofs` references along the whole project, and there are a myriad of different data types that such method can return. Of course, all this data types can be conceptually seen as `Vector{Vector{Int}}`. Besides, such methods, can be traversed in several ways, using a single loop,  two nested loops, with and without caches, etc. 

What it comes into my mind is to define a new Vector of vectors, which is built from another vector of vectors, and two indices, `i`, `j`, and a value `v`, such that the new vector fulfills `x[i][j]==v`. But this is in some sense what `CellDofsWithDofFixed` is doing, so I expect to be missing something.

fverdugo:
@amartinhuertas just a minor thing. Can you add a *temporarily* alias 

```
export FESpaceWithLastDofRemoved
const FESpaceWithLastDofRemoved = FESpaceWithConstantFixed
```

and comment in NEWS.md, deprecated section, that " the name `FESpaceWithLastDofRemoved` has been deprecated in favor of   `FESpaceWithConstantFixed`"

With this, the code is 100% backwards compatible and we can publish your changes via v0.14.1 ASAP so that you can use a registered Gridap version in GridapDistributed.

We need to remember to delete the alias in 0.15.

amartinhuertas:
Sure! But take into account that it is not a single renaming. The interface of the constructor has changed. Do we still follow the deprecated path?

fverdugo:
> Sure! But take into account that it is not a single renaming. The interface of the constructor has changed. Do we still follow the deprecated path?

Then, use `@deprecate` as much as you need like in
https://github.com/gridap/Gridap.jl/blob/5faa6e0742d7003cbae5a7c4bbdefccf08e038c5/src/FESpaces/SingleFieldFESpaces.jl#L172

amartinhuertas:
I was not precise enough in my previous message, sorry about that. 

The new data type has introduced a lot of changes with respect to the previous data type. It has a a new template parameter! I can start using deprecate everywhere, but I have the impression that I am some how maintaining the previous data type alive.  I am not sure it this is what we want. I guess that you want to avoid a new minor relese right? (0.15)

fverdugo:
the old type is a trivial particular case of the new one. We only need these temporary defs until 0.15:

 ```julia
export FESpaceWithLastDofRemoved
const FESpaceWithLastDofRemoved{CS} = FESpaceWithConstantFixed{CS,FixConstant}
@deprecate FESpaceWithLastDofRemoved(space::SingleFieldFESpace) = FESpaceWithConstantFixed(space,true)
```

this 3 lines allow us to publish 0.14.1 ASAP so that you can use it in GridapDistributed

fverdugo:
@amartinhuertas ☝️

fverdugo:
FYI, from the julia doc:

>  As of Julia 1.5, functions defined by @deprecate do not print warning when julia is run without the --depwarn=yes flag set, as the default value of --depwarn option is no. The warnings are printed    from tests run by Pkg.test().

amartinhuertas:
> the old type is a trivial particular case of the new one. We only need these temporary defs until 0.15:

Ok, good call. I was expecting something more dirty to keep the old data type alive. New lesson learned! Done!

amartinhuertas:
I realized that this PR was already merged, perhaps accidentally? I will open a brand new one, I cannot re-open this one.

PR #409, opened 2020-09-16
Added missing get_constant_approach() getter to FESpaceWithConstantFixed

amartinhuertas:


fverdugo (changes requested):
@amartinhuertas 

`get_constant_approach` should return a type or an instance? Normaly, these functons return an instance so that you can do dispatch easily

amartinhuertas:
> get_constant_approach should return a type or an instance? Normaly, these functons return an instance so that you can do dispatch easily

I guess this is a matter of convention, right? 

Using instances you can dispatch as `::T` instead of `::Type{T}`, right? 

Done!

fverdugo:
> Using instances you can dispatch as ::T instead of ::Type{T}, right?

Yes, this is the only reason. The first option is kind of nicer.

fverdugo (approved):

fverdugo:
@amartinhuertas if you want to register a new version, could you please prepare the release in branch release-0.14 (you will need first pull from master `git pull origin master` from `release-0.14`) and open a PR ?

A couple of important things from https://github.com/gridap/Gridap.jl/wiki/How-to-register-a-julia-package

>Make sure that the version number in the Project.toml is the one you want to register (usually version 0.1.0 for new packages).

> Before releasing, make sure that the NEWS.md file contains the right date and version number and that Project.toml has the right version number.

amartinhuertas:
>could you please prepare the release in branch release-0.14 (you will need first pull from master git pull origin master from release-0.14) and open a PR ?

Sure! I will do it now.

Issue #414, opened 2020-10-07
3-tensor contractions with 2-tensor and 1-tensor


Omega-xyZac:
This doesn't seem to be implemented just yet. Would definitely be beneficial.

fverdugo:
Hi @Omega-xyZac,

I thought that these contractions were rather non standard. Do you really need them?

Do you need simple, or full contractions ? which one of the following?

```
a_i = b_ijk*c_jk
a_ijl = b_ijk*c_kl
a_ij = b_ijk*c_k
```

In any case, you can take a look how contractions are implemented in 
https://github.com/gridap/Gridap.jl/blob/master/src/TensorValues/Operations.jl
and try to add the ones you need, but we need to agree in the symbol used to express each case though.

Omega-xyZac:
Hi @fverdugo,

I guess I can use Voigt notation but I would prefer not to since I am working on a homogenisation problem which would ideally have use full tensor formulation. I need:
```
a_i = b_ijk*c_jk
a_ij = b_kij*c_k
```
in the above, `c_jk` would be symmetric. I guess symbol wise we would use `\odot` 2nd case and `\cdot` for first?

fverdugo:
We use \odot for full contraction and \cdot for simple contraction. This would be the syntax for your cases:
```julia
a = b ⊙ c # a_i = b_ijk*c_jk
a = c ⋅ b #  a_ij = c_k*b_kij
```

Omega-xyZac:
Thanks. I'll have a look at implementing those.

Omega-xyZac:
This seems to do the trick:
```
@generated function dot(a::A, b::B) where {A<:MultiValue{Tuple{D1,D2,D3}},B<:MultiValue{Tuple{D1}}} where {D1,D2,D3}
  ss = String[]
  for k in 1:D3
    for j in 1:D2
      s = join([ "a[$i,$j,$k]*b[$i]+" for i in 1:D1])
      push!(ss,s[1:(end-1)]*", ")
    end
  end
  str = join(ss)
  Meta.parse("TensorValue{$D2,$D3}(($str))")
end

import Gridap.TensorValues.inner

@generated function inner(a::A, b::B) where {A<:MultiValue{Tuple{D1,D2,D3}},B<:MultiValue{Tuple{D2,D3}}} where {D1,D2,D3}
  ss = String[]
  for i in 1:D1
[...]

PR #415, opened 2020-10-07
3-Tensor Operations and Constructors

Omega-xyZac:
Added functionality to compute
```
a_ij = b_kij*c_k
a_i = b_ijk*c_jk
```

### DIFF
--- a/src/TensorValues/Operations.jl
+++ b/src/TensorValues/Operations.jl
@@ -274,8 +274,8 @@ const ⊙ = inner
   Meta.parse("VectorValue{$D1}(($str))")
 end
 
-# a_ijpm = b_ijkl*c_klpm
-@generated function double_contraction(a::A, b::B) where {A<:SymFourthOrderTensorValue{D},B<:SymFourthOrderTensorValue{D}} where D
+# a_ijpm = b_ijkl*c_klpm (3D)
+@generated function double_contraction(a::A, b::B) where {A<:SymFourthOrderTensorValue{3},B<:SymFourthOrderTensorValue{3}}
 
   Sym4TensorIndexing = [1111, 1121, 1131, 1122, 1132, 1133, 2111, 2121, 2131, 2122, 2132, 2133,
                         3111, 3121, 3131, 3122, 3132, 3133, 2211, 2221, 2231, 2222, 2232, 2233,
@@ -284,11 +284,32 @@ end
   for off_index in Sym4TensorIndexing
     i = parse(Int,string(off_index)[1]); j = parse(Int,string(off_index)[2]);
     m = parse(Int,string(off_index)[3]); p = parse(Int,string(off_index)[4]);
-    s = join([ "a[$i,$j,$k,$l]*b[$k,$l,$m,$p]+" for k in 1:D for l in 1:D])
+    s = join([ "a[$i,$j,$k,$l]*b[$k,$l,$m,$p]+" for k in 1:3 for l in 1:3])
     push!(ss,s[1:(end-1)]*", ")
   end
   str = join(ss)
-  Meta.parse("SymFourthOrderTensorValue{$D}($str)")
+  Meta.parse("SymFourthOrderTensorValue{3}($str)")
+end
+
+# a_ijpm = b_ijkl*c_klpm (general case)
+@generated function double_contraction(a::SymFourthOrderTensorValue{D}, b::SymFourthOrderTensorValue{D}) where D
+  str = ""
+  for j in 1:D
+    for i in j:D
+      for m in 1:D
+        for p in m:D
+          s = ""
+          for k in 1:D
+            for l in 1:D
+              s *= " a[$i,$j,$k,$l]*b[$k,$l,$p,$m] +"
+            end
+          end
+          str *= s[1:(end-1)]*", "
+        end
+      end
+    end
+  end
+  Meta.parse("SymFourthOrderTensorValue{D}($str)")
 end
 
 # a_ilm = b_ijk*c_jklm
@@ -319,7 +340,7 @@ end
   Meta.parse("TensorValue{$D}($str)")
 end
 
-const ⋅² = ⊡ = double_contraction
+const ⋅² = double_contraction
 
 ###############################################################
 # Reductions
--- a/src/TensorValues/Operations.jl
+++ b/src/TensorValues/Operations.jl
@@ -166,17 +166,30 @@ end
     Meta.parse("TensorValue{$D1,$D2}(($str))")
 end
 
-# a_ij = b_kij*c_k
-@generated function dot(a::A, b::B) where {A<:MultiValue{Tuple{D1,D2,D3}},B<:MultiValue{Tuple{D1}}} where {D1,D2,D3}
+# a_ij = b_ijk*c_k
+@generated function dot(a::A, b::B) where {A<:MultiValue{Tuple{D1,D2,D3}},B<:MultiValue{Tuple{D3}}} where {D1,D2,D3}
   ss = String[]
-  for k in 1:D3
+  for i in 1:D1
     for j in 1:D2
-      s = join([ "a[$i,$j,$k]*b[$i]+" for i in 1:D1])
+      s = join([ "a[$i,$j,$k]*b[$k]+" for k in 1:D3])
       push!(ss,s[1:(end-1)]*", ")
     end
   end
   str = join(ss)
-  Meta.parse("TensorValue{$D2,$D3}(($str))")
+  Meta.parse("TensorValue{$D1,$D2}($str)")
+end
+
+# a_ij = c_k*b_kij
+@generated function dot(a::A, b::B) where {A<:MultiValue{Tuple{D1}},B<:MultiValue{Tuple{D1,D2,D3}}} where {D1,D2,D3}
+  ss = String[]
+  for k in 1:D3
+    for j in 1:D2
+      s = join([ "a[$i]*b[$i,$j,$k]+" for i in 1:D1])
+      push!(ss,s[1:(end-1)]*", ")
+    end
+  end
+  str = join(ss)
+  Meta.parse("TensorValue{$D2,$D3}($str)")
 end
 
 # Double contraction
--- a/src/TensorValues/Operations.jl
+++ b/src/TensorValues/Operations.jl
@@ -595,7 +595,7 @@ for op in (:symmetric_part,)
     end
 end
 
-for op in (:inner,:outer,:⋅¹,:⋅²)#,:(:))
+for op in (:inner,:outer,:⋅²)#,:(:))
     @eval begin
         ($op)(a::GridapType,b::GridapType) = operate($op,a,b)
         ($op)(a::GridapType,b::Number)     = operate($op,a,b)
--- a/src/TensorValues/Operations.jl
+++ b/src/TensorValues/Operations.jl
@@ -207,18 +207,18 @@ end
   Meta.parse("ThirdOrderTensorValue{$D}($str)")
 end
 
-# a_il = b_ij*c_jl
-@generated function dot(a::A,b::B) where {A<:MultiValue{Tuple{D1,D2}},B<:MultiValue{Tuple{D2,D3}}} where {D1,D2,D3}
-  ss = String[]
-  for l in 1:D3
-    for i in 1:D1
-      s = join([ "a[$i,$j]*b[$j,$l]+" for j in 1:D2])
-      push!(ss,s[1:(end-1)]*", ")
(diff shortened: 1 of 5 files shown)


src/TensorValues/Operations.jl
```diff
@@ -166,6 +166,19 @@ end
     Meta.parse("TensorValue{$D1,$D2}(($str))")
 end
 
+# a_ij = b_kij*c_k
+@generated function dot(a::A, b::B) where {A<:MultiValue{Tuple{D1,D2,D3}},B<:MultiValue{Tuple{D1}}} where {D1,D2,D3}
```
fverdugo, line 170: This signature breaks our definition of single contraction. We always contract the last index of the first argument with the first index of the second argument (I.e. the indices that are in "contact"). We have two options:

Either move the c to the first argument
```
a = c ⋅ b #  a_ij = c_k*b_kij
```
or contract `j`
```
a = c ⋅ b #  a_kj = b_kij*c_j
```

Perhaps we want to include both.

fverdugo (changes requested):
@Omega-xyZac thanks for your PR.

I have added some coments to be addressed (see inline).

In addition: 

* Please, add some test for the new functionality in test/TensorValuesTests/OperationsTests.jl
* Before merging this, please clarify what is causing the you reported in issue #414

src/TensorValues/Operations.jl
Omega-xyZac: Sure. No problem. I will include both variations. In regards to the error in #414, do you have any ideas what might cause this?

Omega-xyZac:
Bug reported in #414 is due to not using @law for part of the weak form. Unrelated to new functionality.

Omega-xyZac:
I guess we need a new symbol instead of `:` @fverdugo.

fverdugo:
> I guess we need a new symbol instead of : @fverdugo.

Sure @Omega-xyZac ! Perhaps we can use  ⊡ (\boxdot)

In addition, try to respect our convention in which the indices that are contracted are the ones that are in "contact".  That is: 
```julia
a = b ⊡ c # a_ijkl = b_ijab*c_abkl
```

This convention is important since products with tensors are not always commutative.

Omega-xyZac:
> 
> 
> > I guess we need a new symbol instead of : @fverdugo.
> 
> Sure @Omega-xyZac ! Perhaps we can use ⊡ (\boxdot)
> 
> In addition, try to respect our convention in which the indices that are contracted are the ones that are in "contact". That is:
> 
> ```julia
> a = b ⊡ c # a_ijkl = b_ijab*c_abkl
> ```
> 
> This convention is important since products with tensors are not always commutative.

Ah yes sorry! That was actually a typo in the comment. The new `\boxdot` functionality adheres to that.

santiagobadia:
Why do we introduce a new symbol for the double contraction of tensors? There is a standard symbol for this and it is : .  I would keep : unless there is a very compelling reason.

Omega-xyZac:
> 
> 
> Why do we introduce a new symbol for the double contraction of tensors? There is a standard symbol for this and it is : . I would keep : unless there is a very compelling reason.

I agree but there seems to be issues with `:`? Perhaps @fverdugo can explain this more.

fverdugo:
@santiagobadia the symbol : has a very concrete meaning in julia (range).  I find very confusing to give it another meaning.

santiagobadia:
There is not going to be any clash between these two cases. The problem is that the precedence of : vs +,- is different between these two cases. Probably, this cannot be fixed. That is a compelling reason. So 👍 with the new symbol.

Omega-xyZac:
Hi all,
I have some other contractions that I would like to add. One of these will break the convention `a_ijmp = b_kij*c_kmp`. I don't see any nice way around this. I can just leave this one out and use it in my personal code.

The others are
```
# a_ilm = b_ijk*c_jklm
# a_ilm = b_ij*c_jlm
# a_il = b_ijk*c_jkl
# a_il = b_ij*c_jl
```

santiagobadia:
The last operations are single and double contractions. You can add them and use the agreed symbols. I would not include the one that breaks the rules in Gridap. I don't think this is standard enough. You can define them in your part.

santiagobadia:
We could probably consider also this notation...
```julia
⋅¹,⋅²,⋅³,⋅⁴,... # n-contractions
⋅¹ == ⋅
⋅² == ⊡
```
I guess we will need 4-contractions of 4-tensors, which are usually expressed as `::` or 3-contractions, etc.

What do you think @fverdugo @Omega-xyZac ?

Omega-xyZac:
This is probably the best solution to the problem of higher order contractions.

src/TensorValues/Operations.jl
```diff
@@ -220,8 +246,40 @@ function inner(a::SymFourthOrderTensorValue{D},b::MultiValue{Tuple{D,D}}) where
   inner(a,symmetric_part(b))
 end
 
+# a_i = b_ijk*c_jk
+@generated function inner(a::A, b::B) where {A<:MultiValue{Tuple{D1,D2,D3}},B<:MultiValue{Tuple{D2,D3}}} where {D1,D2,D3}
```
santiagobadia, line 250: This is not an inner product. This is a double contraction, isn't it?

src/TensorValues/Operations.jl
```diff
@@ -220,8 +246,40 @@ function inner(a::SymFourthOrderTensorValue{D},b::MultiValue{Tuple{D,D}}) where
...
+end
+
 const ⊙ = inner
 
+###############################################################
+# Double Contractions w/ products
+###############################################################
+
+# a_ijpm = b_ijkl*c_klpm
+@generated function ⊡(a::A, b::B) where {A<:SymFourthOrderTensorValue{D},B<:SymFourthOrderTensorValue{D}} where D
```
santiagobadia, line 267: I would give a name to this method, e.g., `double_contraction` and then define the symbol, as we do, e.g., for `dot`.

santiagobadia (changes requested):

src/TensorValues/Operations.jl
Omega-xyZac: Yes. I was trying to follow the convention that has been used in the rest of the file. I'll update these with the above notation

Omega-xyZac:
Noticed this in the build doc
```
$ julia --check-bounds=yes --color=yes -e "if VERSION < v\"0.7.0-DEV.5183\"; Pkg.test(\"${JL_PKG}\", coverage=true); else using Pkg; Pkg.test(coverage=true); end"

   Testing Gridap

 Resolving package versions...

WARNING: Method definition dot(A<:(Gridap.TensorValues.MultiValue{Tuple{D1, D2}, T, N, L} where L where N where T), B<:(Gridap.TensorValues.MultiValue{Tuple{D2, D3}, T, N, L} where L where N where T)) where {D1, D2, D3, A<:(Gridap.TensorValues.MultiValue{Tuple{D1, D2}, T, N, L} where L where N where T), B<:(Gridap.TensorValues.MultiValue{Tuple{D2, D3}, T, N, L} where L where N where T)} in module TensorValues at /home/travis/build/gridap/Gridap.jl/src/TensorValues/Operations.jl:158 overwritten at /home/travis/build/gridap/Gridap.jl/src/TensorValues/Operations.jl:212.

  ** incremental compilation may be fatally broken for this module **

WARNING: Method definition dot(Gridap.Helpers.GridapType, Gridap.Helpers.GridapType) in module Helpers at /home/travis/build/gridap/Gridap.jl/src/Helpers/GridapTypes.jl:41 overwritten in module TensorValues at /home/travis/build/gridap/Gridap.jl/src/TensorValues/Operations.jl:600.

  ** incremental compilation may be fatally broken for this module **
```
I think this is due to lines 339,340, and 598-606. Is this the correct way to be doing this? Perhaps `⋅¹` should not be in like 598.

santiagobadia (approved):

src/TensorValues/Operations.jl
```diff
@@ -207,18 +207,18 @@ end
...
+#   ss = String[]
+#   for l in 1:D3
+#     for i in 1:D1
+#       s = join([ "a[$i,$j]*b[$j,$l]+" for j in 1:D2])
+#       push!(ss,s[1:(end-1)]*", ")
+#     end
+#   end
+#   str = join(ss)
+#   Meta.parse("TensorValue{$D1}($str)")
+# end
```
santiagobadia, line 222: The commented method seems wrong. It is supposed to be a single contraction of a 2-tensors but you return a 1-tensor. It should return a 2-tensor, which is the method implemented in L157. I would eliminate this commented part.

santiagobadia (changes requested):

fverdugo:
> We could probably consider also this notation...
> 
> ```julia
> ⋅¹,⋅²,⋅³,⋅⁴,... # n-contractions
> ⋅¹ == ⋅
> ⋅² == ⊡
> ```
> 
> I guess we will need 4-contractions of 4-tensors, which are usually expressed as `::` or 3-contractions, etc.
> 
> What do you think @fverdugo @Omega-xyZac ?

I dont know if these names are going to be interpreted as operators by the julia compiler

src/TensorValues/Operations.jl
Omega-xyZac: I think that because of the number of inputs, it defaults to a 2-tensor. Regardless, the code can be removed. Will do.

Omega-xyZac:
> > We could probably consider also this notation...
> > ```julia
> > ⋅¹,⋅²,⋅³,⋅⁴,... # n-contractions
> > ⋅¹ == ⋅
> > ⋅² == ⊡
> > ```
> > 
> > 
> > I guess we will need 4-contractions of 4-tensors, which are usually expressed as `::` or 3-contractions, etc.
> > What do you think @fverdugo @Omega-xyZac ?
> 
> I dont know if these names are going to be interpreted as operators by the julia compiler

This seems to be working so far in my testing and other code.

santiagobadia (approved):

Omega-xyZac:
#416 addressed here.

src/TensorValues/Operations.jl
```diff
@@ -222,6 +259,68 @@ end
...
+    push!(ss,s[1:(end-1)]*", ")
+  end
+  str = join(ss)
+  Meta.parse("VectorValue{$D1}(($str))")
+end
+
+# a_ijpm = b_ijkl*c_klpm
+@generated function double_contraction(a::A, b::B) where {A<:SymFourthOrderTensorValue{D},B<:SymFourthOrderTensorValue{D}} where D
+
+  Sym4TensorIndexing = [1111, 1121, 1131, 1122, 1132, 1133, 2111, 2121, 2131, 2122, 2132, 2133,
```
fverdugo, line 280: This is particular for 3D. Can you implement it in general?
If for performance reasons, you want to keep the 3D specialization then use the signature 

```julia
function double_contraction(a::A, b::B) where {A<:SymFourthOrderTensorValue{3},B<:SymFourthOrderTensorValue{3}}
```

fverdugo (changes requested):
I have added some more comments (see inline)

Note that ` ⋅² ` and  `⊡` are the same. I would only keep one version, perhaps ` ⋅² `. On the other hand, having `double_contraction`  is fine since it is the plain asci version.

src/TensorValues/Operations.jl
Omega-xyZac: Okay, perhaps `⋅²`. I will keep the 3D version since this should be fairly quick. I will have to have a think about how to do this generally.

Omega-xyZac: This might do the trick:
```
@generated function double_contraction(a::SymFourthOrderTensorValue{D}, b::SymFourthOrderTensorValue{D}) where D
  str = ""
  for j in 1:D
    for i in j:D
      for m in 1:D
        for p in m:D
          s = ""
          for k in 1:D
            for l in 1:D
              s *= " a[$i,$j,$k,$l]*b[$k,$l,$p,$m] +"
            end
          end
          str *= s[1:(end-1)]*", "
        end
      end
    end
  end
  Meta.parse("SymFourthOrderTensorValue{D}($str)")
end
```

Need to test.

santiagobadia: Yes, keep ⋅² so we use the same notation for all >1 contractions.

fverdugo (approved):

fverdugo:
@Omega-xyZac 

can you include some comment on the new functionality in the https://github.com/gridap/Gridap.jl/blob/master/NEWS.md file ?

Please add the comments in a new section called `[Unreleased]`, i.e.:

```
## [Unreleased]

## Added
- Your commend there...
```

Thanks!

PR #417, opened 2020-10-14
Bugs detected in refactored code (some of them fixed, others no)

fverdugo:


fverdugo:
@santiagobadia 

I have found several important bugs. This PR solves some of them (but not all). We need to discuss face-to-face since there is one bug that it is quite fundamental (I know how to solve it, but It would requires some work. So, I prefer to talk first)

I have included tests that trigger these bugs to avoid regressions in the future. The tests associated with the bugs that are not fixed are marked as `@test_broken`

PR #418, opened 2020-10-14
Removing inference module

fverdugo:


fverdugo:
@santiagobadia 

In this PR I have fixed the pending issues from previous one (PR #417)

At the end, the solution was as "simple" as removing the inference module. This module is not needed anymore and it was causing problems. So, I have removed it. 

We don't need the Inference module any more since one can use the machinery related with function `return_type` implemented at `Map` level. `Function` is just a particular case. Thus, we don't need a dedicated module for inferring return types from functions.

Issue #444, opened 2020-11-06
Replace hard-coded `Int64` by `Int` wherever appropriate


amartinhuertas:
Using a hard-coded `Int64` instead of `Int` in the code can result in Julia dispatching errors  with the 32-bit version of Julia (in which `Int` means `Int32`) due to mixed precision integers (Int64 versus Int32) not being accounted for in type parameterized functions.

See, e.g., https://github.com/gridap/GridapDistributed.jl/runs/1362528777#step:5:87, with Gridap v0.14.1

amartinhuertas:
Solved in #445. Closing the issue.

PR #445, opened 2020-11-06
Replace int64 by int

amartinhuertas:
Related to issue https://github.com/gridap/Gridap.jl/issues/444

Note: Using temporary Github actions ci script. We can convert this into the definitive one in a separate PR. 

What do u think @fverdugo ?


fverdugo (approved):
@amartinhuertas thanks for the PR!

I merge this as it is. We can improve the ci script in a separate PR.

PR #456, opened 2020-11-17
Adding x86 to gh actions

amartinhuertas:


fverdugo (approved):

fverdugo:
@amartinhuertas perhaps  we could run with x86 only in master and in PR against master. 

The test suite is quite heavy and running x86 in all work branches is perhaps too much...

What do you think?

amartinhuertas:
> @amartinhuertas perhaps we could run with x86 only in master and in PR against master.
> 
> The test suite is quite heavy and running x86 in all work branches is perhaps too much...
> 
> What do you think?

Yes, I agree, no need to perform x86 on all branches. I will do it.

amartinhuertas:
>Yes, I agree, no need to perform x86 on all branches. I will do it.

Done in https://github.com/gridap/Gridap.jl/pull/461!

Issue #408, opened 2020-09-15
delta functions?


stevengj:
It would be nice to be able to represent Dirac delta distributions in order to solve a PDE with a delta-function right-hand-side.

Basically each delta-function term δ(x-x₀) would reduce the dimensionality of the integration (for assembly) by 1, but the implementation requires more understanding of the `CellField` mechanisms than we currently have.

cc @wenjie

fverdugo:
Hi @stevengj!

yes, this is in our TO-DO list.

I am working in an enhanced notation for weak forms (see issue #265). Once this is available one could implement something like:

```julia
f(x) = x[1]
x0 = Point(1.4, 3.0)
l(v)  = ∫( f*v )*dΩ + 3.5*v(x0)
```

or 

```julia
l(v)  = ∫( f*v  +  3.5*δ(x0) )*dΩ
```

stevengj:
That sounds great.  Ideally, we'd also want surface delta functions, the simplest case of which is a delta function of just one or two of the cartesian coordinates. e.g.
```jl
l(v) = ∫ (f*v)*δ(x[1]-x0[1]) )*dΩ
```
(For example, to implement mode-launching sources in wave equations.)

fverdugo:
@santiagobadia @stevengj 

I have added (PR #485) `DiracDelta` to impose point, line, or surface loads. The usage is pretty straightforward.

```julia
degree = 2
δ_point = DiracDelta{0}(model,tags=["mypoint1","mypoint2"])
δ_line = DiracDelta{1}(model,degree,tags="myline")
δ_surf = DiracDelta{2}(model,degree,tags="mysurf")
```
For point loads `degree` is optional. For line/surface loads `degree` is used to choose the quadrature rule. The implementation is efficient, since we relay on physical tags defined in the model (we do not need to look for which cell includes a point)

Once defined, it is used in the linear form together with other terms:
```julia
l(v) = δ_point(v) + # other terms
```

Example: linear elasticity with point + body forces:

```julia
using Gridap

domain = (0,1,0,1)
cells = (30,30)
model = CartesianDiscreteModel(domain,cells)

Ω = Triangulation(model)

order = 2
reffe = ReferenceFE(:Lagrangian,VectorValue{2,Float64},order)
V = FESpace(model,reffe,dirichlet_tags=[1,2,5])

dΩ = LebesgueMeasure(Ω,2*order)

δ = DiracDelta{0}(model,tags=4)

E = 1
[...]

PR #485, opened 2020-11-27
Added DiracDelta

fverdugo:
Fixes issue #408


stevengj:
How do you specify the amplitude?  Can you do e.g. `3 * DiracDelta(...)` or `f * DiracDelta` (where `f(x)` is a function of position)?

stevengj:
What do you do if you have piecewise-constant test functions, since then you have a discontinuity right where the delta function is?  I guess you could use the average of the two sides…

PR #486, opened 2020-11-27
Added ∑ symbol for the `sum` function

oriolcg:
I'm not completely sure about the location of the definition of ∑. Since its a quite generic operator, I added it in the general file `Arrays.jl` .

I also added some tests for different types of sums.


fverdugo:
@santiagobadia do you like to use `∑` as an alias of `sum`?

At the begging I liked the idea, but now I am not sure, perhaps is too much.

santiagobadia:
@fverdugo I don't think it is that important, but it doesn't harm either. So, we can add it to Gridap, it is just a couple of lines.

PR #496, opened 2020-12-08
Number domain contribution

oriolcg:
This PR is required to enable products of `Number` against `DomainContributions`. This is particularly needed in GridapODEs, where the transient jacobian is multiplied by the time step factor.


src/CellData/DomainContributions.jl
```diff
@@ -38,14 +38,14 @@ function add_contribution!(a::DomainContribution,trian::Triangulation,b::Abstrac
   if length(a.dict) > 0
     T = eltype(first(values(a.dict)))
     if T <: AbstractMatrix
-      @assert S<:AbstractMatrix """\n
+      @assert (S<:AbstractMatrix || (S<:Number && op == *)) """\n
```
fverdugo, line 41: I find a bit dangerous to relax these checks in that way.

src/CellData/DomainContributions.jl
```diff
@@ -93,6 +93,18 @@ function (-)(a::DomainContribution,b::DomainContribution)
...
 
+function (*)(a::Number,b::DomainContribution)
+  c = copy(b)
+  for (trian,array) in b.dict
+    s = size(get_cell_map(trian))
+    constant_array = Fill(a,s)
+    add_contribution!(c,trian,constant_array,*)
+  end
+  c
+end
```
fverdugo, line 104: I would implement this in a different way so that you do not need to relax the checks above. Something like:

```julia
function (*)(a::Number,b::DomainContribution)
  c = DomainContribution()
  for (trian,array_old) in b.dict
    s = size(get_cell_map(trian))
    array_new = lazy_map(Broadcast(*),Fill(a,s),array_old)
    add_contribution!(c,trian,array_new)
  end
  c
end
```

fverdugo (changes requested):
@oriolcg thanks for the PR!

Please, consider my comments below.

src/CellData/DomainContributions.jl
oriolcg: Done in 05c884e45c9bddac2cb2db061e8fe3fcf882c9af

fverdugo (approved):

PR #506, opened 2020-12-14
Lazy array renaming

santiagobadia:


santiagobadia:
Given a `a::LazyArray`, we rename 

`a.g` -> `a.maps`
`a.f` -> `a.args`

PR #514, opened 2021-01-11
Laplacian

oriolcg:
This pull request enables the computation of Laplacians. Currently only available for FE with `AffineMap`.

### DIFF
--- a/test/GridapTests/BiharmonicTests.jl
+++ b/test/GridapTests/BiharmonicTests.jl
@@ -1,0 +1,54 @@
+module BiharmonicTests
+
+using Test
+using Gridap
+
+# Analytical manufactured solution
+u(x) = x[1]*(x[1]-1)*x[2]*(x[2]-1)
+f(x) = Δ(Δ(u))(x)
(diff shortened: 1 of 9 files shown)


test/GridapTests/BiharmonicTests.jl
```diff
@@ -0,0 +1,54 @@
+module BiharmonicTests
```
fverdugo, line 1: Is this file being run by the test suite?

It seems that it needs to be included into `test/GridapTests/runtests.jl`

fverdugo (changes requested):
Hi @oriolcg, ethanks for the PR!

Can you also add a line into NEWS.md (###Added section)

See my other comment below.

test/GridapTests/BiharmonicTests.jl
oriolcg: Done in 2b7c17090a5681cbb34477681087c0f8b0bab04c

fverdugo (approved):

PR #519, opened 2021-01-22
Provide simplexify for arbitrary dimensions

eschnett:


fverdugo (changes requested):
Hi! @eschnett 

Thanks for the PR. It is very nice to have this general formula.

In some places in Gridap (e.g. for Raviart-Thomas FEs), we need that the connectivity of the generated simplices are ordered. I.e., the node ids of a given simplex are ordered from small to large.

Is this property, fulfilled in your general formula?

eschnett:
What do you mean by "ordered"? Do you refer to the way the vertices within each simplex are ordered? If so, they should be; I can add a test as well. Or do you refer to a particular ordering of the simplices the list of generated simplices?

fverdugo:
> What do you mean by "ordered"? Do you refer to the way the vertices within each simplex are ordered? If so, they should be; I can add a test as well. Or do you refer to a particular ordering of the simplices the list of generated simplices?

Vertices within each simplex. The vertex ids of a given simplex need to be ordered from small to large. Adding a test that checks this condition wold be nice.

santiagobadia:
@fverdugo @eschnett have you checked whether the resulting mesh is oriented?

@eschnett what we exploit in Gridap (and some other FE packages) is the fact that simplicial meshes can always be edge-oriented. It means that the local numbering of nodes of a simplex and the global numbering of nodes in the mesh is such that if you take any edge of the mesh, the orientation of the edge determined by the global vertices Id is identical to the one using the local vertex numbering of any of the cells that contain that edge. You can easily attain this via an obvious postprocess if this is not the case in your current implementation.

eschnett:
@santiagobadia I assume the respective condition is that the local vertex ids are sorted. In this case, yes, they are, and there is a test for this property.

Issue #425, opened 2020-10-20
Make FEFunction objects callable (as functions of coordinates)


stevengj:
Given an `f::FEFunction`, it would be nice to be able to evaluate it at a position `x::VectorValue` just by calling `f(x)`.   (Right now this is a `MethodError`)

(Presumably you have this functionality already, and this is just a one-line method addition?)

santiagobadia:
Not really, this issue is very related to the Dirac delta one.

A FE function is a cell-wise function. To evaluate `f(x)` we need first to find which is the cell in the mesh that contains `x`. (This value can be undefined when is x is on the skeleton mesh and the space is not C0, which is the case of Nedelec, Raviart-Thomas, etc). This is something we do not use because it is expensive and we do not need it in 99% of cases. All the time, our points are already defined in a cell (integration points) and we can skip that part.

But this functionality, i.e., to define a FE function in an arbitrary point, is needed, e.g., when we have a Dirac delta as forcing term.

Once we have defined the algorithm such that given `x` returns _a_ cell in the mesh such that its closure contains `x` for the different `Gridap` meshes, the implementation will be straightforward. That is the missing ingredient. The implementation for Cartesian mesh should be pretty simple if this is what you are using. But I think you were interested on unstructured meshes, from what I read in the Nedelec issue.

stevengj:
> This is something we do not use because it is expensive and we do not need it in 99% of cases. All the time, our points are already defined in a cell (integration points) and we can skip that part.

I agree that you don't need it during assembly etcetera, but the ability to evaluate the solution is pretty convenient for users doing post-processing and visualization.   (e.g. in our case, we wanted to validate our solution by comparing it to a known solution at some set of points.)

(It would also be useful if you needed to interpolate a solution from one mesh to another; e.g. Fenics supports this IIRC.)

To implement it reasonably efficiently (with log(N) complexity) you'd need a tree of some sort, e.g. a kd tree, of course.

santiagobadia:
I agree, I also consider that it is something to be done. But the point to cell map is the part that prevents us to easily provide this functionality now. There is already a k-d-tree searching algorithm implemented in Julia, this [package](https://github.com/JuliaGeometry/KDTrees.jl) that I would try to use here.

eschnett:
[...]

PR #523, opened 2021-01-26
Allow evaluating FE functions at arbitrary points

eschnett:
I'm looking for feedback, in particular:

- I'm using my existing package [`Bernstein.jl`](https://github.com/eschnett/Bernstein.jl), but only call a single function in it. Should I instead import the respective code?
- What mechanism should I use to cache the calculated `kdtree`? Can I somehow attach it to the triangulation?
- Where is @ericneiva's branch that is mentioned in https://github.com/gridap/Gridap.jl/issues/425?
- Did I overlook some existing functionality that would simplify my code?

I'm aware that the code is still quite restrictive (only handles `SingleFieldFEFunction`s with simplicial meshes), and that it uses linear interpolation instead of actually evaluating the basis functions.

Will eventually close https://github.com/gridap/Gridap.jl/issues/425, I hope.


ericneiva:
Hi, @eschnett,

> Where is @ericneiva's branch that is mentioned in #425?

I have answered you in #425. Cheers!

fverdugo:
> I'm using my existing package Bernstein.jl, but only call a single function in it. Should I instead import the respective code?

Which function do you need? Perhaps a gridap equivalent is already there.

> What mechanism should I use to cache the calculated kdtree? Can I somehow attach it to the triangulation?

I would use the caching mechanism we have in gridap.

I.e., implement `return_cache(f::CellField,x::Point)` function. The result of this function, is then the first argument in `evaluate!(cache,f::CellField,x::Point)`. Once these two functions are working you can efficiently evaluate the CellField in a vector of points by creating the cache only once:

```julia
points = ... # Vector{Point{D,Float64}}
uh = ... # CellField
cache = return_cache(uh,testitem(points)) # Generate the cache. i.e., do the setup
for x in points
  uh_at_x = evaluate!(cache,uh,x)
end
```

> I'm aware that the code is still quite restrictive (only handles SingleFieldFEFunctions with simplicial meshes), and that it uses linear interpolation instead of actually evaluating the basis functions.

Yes, the final implementation should only rely on the CellField interface.

eschnett:
The function I'm using converts a point in a simplex from Cartesian to barycentric coordinates so that I know whether it is inside the simplex or not. This is necessary because the k-d tree search returns the nearest vertex to a give point, and I then have to search all neighbouring simplices. Is there a function for contained-ness?

fverdugo:
Perhaps we can consider this package

https://github.com/JuliaGeometry/GeometricalPredicates.jl

 It has no dependencies, so including it is not harmful

eschnett:
That package only supports 2d and 3d simplices. I'll copy the respective code – it's not much, essentially just constructing and inverting a small matrix.

src/CellData/CellFields.jl
```diff
@@ -204,14 +204,97 @@ DomainStyle(::Type{GenericCellField{DS}}) where DS = DS()
...
+  kdtree = KDTree(map(nc -> SVector(nc.data), node_coordinates))
+  kdtree::KDTree{SVector{D,T},Euclidean,T}
+  return CellFieldCache(kdtree)
+end
+
 function evaluate!(cache,f::CellField,x::Point)
-  @notimplemented """\n
-  Evaluation of a CellField at a given Point is not implemented yet.
+  cache::CellFieldCache
+
```
santiagobadia, line 227: We could probably encapsulate this code in a function and do dispatching with respect to the triangulation type. The code below is for an unstructured (simplicial) mesh. We could also consider the case for a structured Cartesian mesh (which is much easier, no tree involved).

src/CellData/CellFields.jl
```diff
@@ -204,14 +204,97 @@ DomainStyle(::Type{GenericCellField{DS}}) where DS = DS()
...
+      push!(dists, dist)
+    end
+
+    dist, m = findmax(dists)
+    cell = cells[m]
+    λ = bary(cell, x)
+
+    # Ensure the point is inside one of the cells, up to round-off errors
+    @assert dist ≥ -1000 * eps(T) "Point is not inside a cell"
+
```
santiagobadia, line 289: We have to decide what happens if the point is in more than once cell. We can provide just one of the cells. The result is going to be identical for C0 spaces. For discontinuous spaces, the query is ill-posed (the function is not defined at this point) and we would just return the function in one of the cells.

santiagobadia:
> I'm using my existing package Bernstein.jl, but only call a single function in it. Should I instead import the respective code?

Since this is unrelated to Berstein polynomials I would rather add the code here (I guess it is not much)

> Is there a function for contained-ness?

No, I am not sure what is the best geometrical predicate for determining whether a a point belongs to an n-simplex. The one you have implemented seems a good option.

src/CellData/CellFields.jl
eschnett: I am calculating the distance from the point to all the cells. Without round-off, and with non-overlapping cells, the distance would be positive for exactly one cell and negative for all other ones. (I should probably change my sign convention.) Due to round-off, the distance can be slightly positive or slightly negative for points near cell boundaries, in particular near vertices. In this case, I choose the cell with the largest distance, and I check that the distance (if negative) is at most at round-off level.

I'll update the comments in the code.

santiagobadia: I think that there is the (unlikely) possibility that the distance will be zero (or in the round-off error) for some cells, am I wrong? I was thinking of cells as closed domains, which is the standard definition in finite elements. If you consider that cells are open domains, you can find points that do not belong to any cell. At the end is just `<` vs `<=`, etc.

santiagobadia: In any case, what you propose is resilient to this, so keep it as it is

eschnett: I view the cells as closed as well, and would be happy to return a value from any one of the cells if the point lies on the boundary. The problem I'm addressing here is only round-off error, which is a real, practical problem. For example, asking for the value at a vertex can yield distances that say the point is outside of all the neighbouring cells.

For the cases I tested, this error can be more than 100 eps, but was always less than 1000 eps. This distance is in terms of the barycentric coordinates which have a range of 0...1. In this case, the code will extrapolate from one of the cells by that distance, i.e. about 1e-12 of a cell width.

fverdugo:
You provably also need PR #550

eschnett:
This code works now, for both simplicial and hexahedral grids.

I'm not sure it's efficient. I'd be happy about feedback for improving its efficiency.

eschnett:
As mentioned above, the code works now as intended.

test/FieldsTests/runtests.jl
```diff
@@ -20,4 +20,6 @@ using Test
 
 @testset "InverseFields" begin include("InverseFieldsTests.jl") end
 
+@testset "InverseFields" begin include("InverseFieldsTests.jl") end
```
fverdugo, line 23: Remove this line, so that `InverseFieldsTests.jl` is included only once

src/MultiField/MultiFieldCellFields.jl
```diff
@@ -39,3 +39,167 @@ num_fields(a::MultiFieldCellField) = length(a.single_fields)
 Base.getindex(a::MultiFieldCellField,i::Integer) = a.single_fields[i]
 Base.iterate(a::MultiFieldCellField)  = iterate(a.single_fields)
 Base.iterate(a::MultiFieldCellField,state)  = iterate(a.single_fields,state)
+
```
fverdugo, line 42: Why do you need to implement this here in `src/MultiField/MultiFieldCellFields.jl` ? I would say this should go in `src/CellData/CellFields.jl`. The implementation of the evaluation of a `CellField` in some `Points` is conceptually decoupled from FE spaces,multifield etc. It just needs the `CellField` interface and some of the upstream modules.

test/MultiFieldTests/MultiFieldCellFieldsTests.jl
```diff
@@ -116,6 +118,50 @@ cellmat1_Γ = integrate(((n⋅dv.⁺)-dq.⁻)*((n⋅du.⁺)+dp.⁻),quad_Γ)
 cellmat2_Γ = integrate((n⋅dv.⁺)*(n⋅du.⁺)+(n⋅dv.⁺)*dp.⁻-dq.⁻*(n⋅du.⁺)-dq.⁻*dp.⁻,quad_Γ)
 test_array(cellmat1_Γ,cellmat2_Γ,≈)
 
+# Test function evaluation
```
fverdugo, line 121: As with the implementation, I would move the tests into `test/CellDataTests/CellFieldsTests.jl`

fverdugo (changes requested):
Hi @eschnett. Thanks for the PR!

Please, see my comments below 👇

santiagobadia:
Hi @fverdugo @eschnett 

@balaje is updating @eschnett with master since he needs something similar for his GSoC project. @balaje can do this work and update the PR, ok?

eschnett:
@santiagobadia Are you volunteering @Balaje to update the PR? Let me know if you need help or get stuck. I'd be happy to do this work myself.

Balaje:
Hi @eschnett 

I have already made a PR to your repo with the suggested changes. I see that you have merged the PR. 

@santiagobadia @ericneiva  I have updated `test/CellDataTests/CellFieldsTests.jl` with some more tests. Should I submit another pull request to @eschnett's branch?

santiagobadia:
Hi @Balaje 

Yes, if you make changes in @eschnett fork, please send a PR to @eschnett repo. When @eschnett accepts the PR, this one updates automatically, I think.

Hopefully, we are close to accept the PR. Sorry @eschnett for the mess. Thanks @Balaje for the work.

src/MultiField/MultiField.jl
```diff
@@ -15,12 +15,21 @@ using Gridap.TensorValues
 using Gridap.CellData
 using Gridap.Fields
 
+using Gridap.FESpaces: FEBasis, TestBasis, TrialBasis, get_cell_dof_values
```
fverdugo, line 18: I think the changes in this file are no longer needed after moving the code to `CellFields.jl`. Can you revert them?

src/MultiField/MultiFieldCellFields.jl
```diff
@@ -39,3 +38,5 @@ num_fields(a::MultiFieldCellField) = length(a.single_fields)
...
+# Evaluation of CellFields
```
fverdugo, line 42: Remove this line

fverdugo (changes requested):
Hi @eschnett @Balaje ! Thanks for the changes. I still have a couple of minor comments.

src/CellData/CellData.jl
```diff
@@ -57,6 +61,10 @@ export CellDof
 export get_normal_vector
 export get_cell_measure
 
+export make_inverse_table
+export point_to_cell!
```
fverdugo, line 65: You are exporting `point_to_cell!` but the function to generate its cache object seems to be private.

Here, there are 2 options:

1. Do not export `point_to_cell!` and rename it to  `_point_to_cell!` so that it is clear that it's private.
2.  Make public the also the mechanism to generate the cache. In this case, follow the API of the abstract type `Map` just to be consistent with other parts of the code. 
```julia
struct PointToCellMap{T<:Triangulation} <: Map
  trian::T
end

return_cache(k::PointToCellMap,x::Point) = ...
evaluate!(cache,k::PointToCellMap,x::Point) = ...
```

fverdugo (changes requested):
Hi @Balaje @eschnett !

Still some more minor comments.

santiagobadia:
I have created an issue with the last minor comment. I proceed to merge this PR.

Issue #527, opened 2021-01-26
Why PR filed by CompatHelper do not trigger the CI pipeline?


fverdugo:
@amartinhuertas can you help with this ?

amartinhuertas:
> @amartinhuertas can you help with this ?

Yes, sure. I will come back to you when I have a solution ...

amartinhuertas:
It seems that `CompatHelper` is not triggering the CI pipeline because there is a missing `COMPATHELPER_PRIV` record in the `CompatHelper` GitHub action `yml` file. Directly quoting from:

https://github.com/JuliaRegistries/CompatHelper.jl#12-set-up-the-ssh-deploy-key-optional

> The default CompatHelper setup has one flaw: the pull requests that CompatHelper opens will not trigger any other GitHub Actions.

> Consider the following situations:
> You use GitHub Actions to test your package using continuous integration (CI).
> You use GitHub Actions to build and deploy the documentation for your package.
> If any of those situations apply to you, then you will need to set up an SSH deploy key for CompatHelper. Once you have set up an SSH deploy key for CompatHelper, the pull requests that CompatHelper opens will trigger all of the usual GitHub Actions.

Fixing it ...

amartinhuertas:
Handling this issue in PR #528

amartinhuertas:
It seems that CompatHelper is already triggering the CI pipeline, see, e.g., https://github.com/gridap/Gridap.jl/pull/540

Closing the issue ...

PR #528, opened 2021-01-27
Added ssh secret to compathelper yml file

amartinhuertas:
Tries to solve issue #527


amartinhuertas:
@fverdugo ... is there any way to manually trigger CompatHelper from this branch so that we can double check that this PR solves the issue? I did not come up with a solution after quick searching, perhaps we can just accept the PR and then keep in mind that we have to check this in the next CompatHelper action.

fverdugo:
> we can just accept the PR and then keep in mind that we have to check this in the next CompatHelper action.

Let's do this!

PR #532, opened 2021-01-28
simplexify: return tet simplex instead of hex simplex in 1d

eschnett:


eschnett:
Currently, 1d grids are always a `SEGMENT`, which is internally a `HEX` grid. This patch introduces a distinction between 1d `TET` and `HEX` grids, i.e. between 1d cubes and 1d simplices. The advantage is that `simplexify` now always returns simplices, not cubes.

Formally this makes sense to me, practically there should be no difference. Let me know if you agree, or if you want to drop this patch.

fverdugo:
It makes sense, thanks!

PR #537, opened 2021-02-02
Default implementation of compress(cell_mat,trian)

pmartorell:
Default dummy implementation of `compress() ` which will allow to compress the cell matrices before the assembling. Thus, reduce the memory consumption when multiple cell matrices point to the same cell, e.g., GridapEmbedded.


fverdugo:
@pmartorell Thnaks for creating the branch. I close the PR so that we can work on this branch without triggering so many GH actions. We can open it again once the work is done.

PR #538, opened 2021-02-02
Add norm operation for TensorValues

Paulms:
In the p-Laplacian tutorial you can use norm∘∇(u). If you try to do the same: norm∘T where T belongs to a Tensor Valued Space (like, TestFESpace(model,ReferenceFE(lagrangian,TensorValue{2,2,Float64,4},order),conformity=:L2)), it doesn`t work. Adding the method: norm(u::MultiValue{Tuple{D1,D2}}) where {D1,D2} = sqrt(inner(u,u)) in TensorValues/Operations.jl seems to solve the issue.


fverdugo:
Thanks for the PR!

PR #542, opened 2021-02-09
Adding a new method for dot function for third order tensor values.

diliphridya:


src/TensorValues/Operations.jl
```diff
@@ -207,6 +207,20 @@ end
   Meta.parse("ThirdOrderTensorValue{$D}($str)")
 end
 
+@generated function dot(a::A,b::B) where {A<:MultiValue{Tuple{D,D}},B<:ThirdOrderTensorValue{D,D,L}} where {D,L}
```
fverdugo, line 210: Delete previous method (line 196) since it is not needed any more.

fverdugo (changes requested):
Hi @diliphridya. Thanks for the PR.

Some test are nor passing. Please take a look.

Se also my in-lined comment.

src/TensorValues/Operations.jl
diliphridya: Sure. I will take a look and make the changes.

PR #557, opened 2021-03-06
Provide InverseField

eschnett:
An InverseField provides the inverse of a Field, calculated on the fly via NLsolve.


src/Fields/FieldArrays.jl
```diff
@@ -664,3 +664,37 @@ for op in (:*,:⋅,:⊙,:⊗)
...
+function return_cache(a::InverseField,x::Point)
+  return return_cache(a.original,x), return_cache(∇(a.original),x)
+end
+
+function evaluate!(caches,a::InverseField,x::Point)
+  P = typeof(x)
+  D = length(x)
+  cache,∇cache = caches
+  # Initial guess. (Can this be improved?)
+  y₀ = zero([x...])
```
fverdugo, line 686: Use this to avoid allocations: y0 = SVector(Tuple((zero(P))))

fverdugo (changes requested):
@eschnett Thanks for the PR! see comments below plus this ones:
- Move code to file `src/Fields/InverseFields.jl`
- Write some test is `test/FieldsTests/InverseFieldsTests.jl`

src/Fields/FieldArrays.jl
eschnett: Unfortunately, NLsolve requires a mutable vector, and MVector does not work. We're stuck with a regular array.

I couldn't find any good alternatives to NLsolve either.

eschnett:
I moved the code and added tests.

src/Fields/FieldArrays.jl
fverdugo: This should work (at least in the Gridap part), if not i am overlooking something.
```julia
  y₀ = MVector(Tuple(zero(P)))
  f!(F,y) = F .= SVector{D}(Tuple(evaluate!(cache,a.original,P(Tuple(y))) - x))
  j!(J,y) = J .= SMatrix{D,D}(Tuple(evaluate!(∇cache,∇(a.original),P(Tuple(y)))))
```

eschnett: Using an `MVector` leads to a nan in NLsolve at run time, for a very simple test case with an affine function. This nan then makes the solver abort without converging.

Using a Julia array does not encounter a nan. I assume that NLsolver doesn't handle `MVector`s correctly for some reason.

fverdugo: OK! Then allocate the vector `y0` in the cache.

eschnett: Done. I also allocate `F0` in the cache.

fverdugo (approved):

PR #574, opened 2021-04-01
Avoid a call to _RT_nodes_and_moments in return_cache

amartinhuertas:
To-think: should we do a `deepcopy` of the nodes and moments arrays before creating the MomentBasedDofBasis?


amartinhuertas:
Tests are not passing after the change. I will investigate the issue and come back whenever I have some time.

amartinhuertas:
> To-think: should we do a deepcopy of the nodes and moments arrays before creating the MomentBasedDofBasis?

I do not think so. At least the tests pass without deep copy.

> Tests are not passing after the change. I will investigate the issue and come back whenever I have some time.

Fixed. PR ready to be reviewed. In this PR we just avoid calling `_RT_nodes_and_moments` each time that we create the cache required to evaluate the cell array of RT shape functions on a set of points.

Issue #571, opened 2021-03-24
RT FEs on manifolds (writing in progress)


amartinhuertas:
Note: Some connection with issue #500 

The implementation of RT FEs on manifolds presents some subtle differents with respect to the case of matching topological and geometrical dimensions. As my understanding improves, I will try to document here what is needed to support these. Work in progress ...

1. Direct quotation from Rognes et. al. paper (**Automating the solution of PDEs on the sphere and other manifolds in FEniCS 1.2**). "However, the finite element families mapped via the contravariant Piola transform present one additional novel aspect: the choice of sign in Eq. (30). As detailed in Rognes et al. (2009) for the case of matching topological and geometric dimensions, using the signed determinant of the Jacobian in combination with the UFC numbering of mesh entities result in the desired normal component continuity. **However, as the generalised Jacobian determinant does not carry a sign, the choice of sign in Eq. (30) must be determined by alternative means.** In fact, it is not possible to determine the sign based on the geometry of the physical cell only. Therefore, the code generation assumes that the choice of sign is determined via an additional input argument to the relevant UFC functions."
2. In our case, the triangulation of the manifold should provide whether each cell (e.g., triangle in 3D) was oriented (negative sign) during mesh generation or not (positive sign). We should be able to extract this information out of the triangulation and provide it to the functions dealing with the implementation of the `ContraVariantPiolaMap`.

amartinhuertas:
Solve in PR #577 and #579

PR #577, opened 2021-04-11
Misc changes in order to support RT FEs in general manifolds (i.e., num_cell_dims(model) < num_point_dims(model))

amartinhuertas:
Hi @santiagobadia, @fverdugo,

In this PR I include a first proposal for a set of miscellaneous changes/fixes required in order to support RT FEs posed over general manifolds that I want to be subject to review/discussion. 

These changes are related, among others, to the remarks given in issue https://github.com/gridap/Gridap.jl/issues/571 and references therein. 

As far as I understand, the model should eventually provide the `cell_orientation` array.  We should also discuss how can we include this info in the abstract interface of `DiscreteModel` and related data types, and how this information is computed in the case of general, unstructured, oriented manifolds. 

Apart from using `cell_orientation` for the implementation of the sign of the ContraVariant Piola Map, I am wondering where (if at any place) this info should also be used. For example, should it be used for the evaluation of dofs, and in particular, for the computation of normals in physical space from those of reference space? I dont think so. However, I am concerned that, in FENICS, this info seems to be required for the evaluation of dofs (see, e.g., https://fenics-dolfin.readthedocs.io/en/2017.2.0/apis/api_ufc.html#_CPPv2NK3ufc14finite_element3ufc14finite_element13evaluate_dofsEPdRK8functionPKdiRK4cellPKN3ufc18coordinate_mappingE) apart from a myriad of other subroutines. Am I missing something?

Currently I am assuming that all cells are oriented as the global manifold in the case of `num_cell_dims(model) < num_point_dims(model)`, i.e., `cell_orientation=false` for all cells. In the case of `num_cell_dims(model) == num_point_dims(model)` I am extracting this info from the sign of the determinant of the Jacobian. Regarding this last point, although it is temporary,  I am concerned about the fact that I believe that we are computing the Jacobian and the determinant for both the ContraVariantPiolaMap and cell_orientation. Is that true? If yes, is there anyway to waive this duplicated computation? On the other hand, the Jacobian and its (pseudo)determinant are also required for the transformation of the integral differential measure (apart from the Piola transformation). Does Gridap ensure that this is only computed once? If yes, which is the mechanism by which it is achieving this?


amartinhuertas:
Tests not passing yet. I have to investigate the cause ...

amartinhuertas:
> Tests not passing yet. I have to investigate the cause ...

Ok. I figure out the cause. The tests which are failing are those corresponding to (oriented) simplices meshes. 

The previous code was using the signed determinant when computing the face RT DoFs. See https://github.com/gridap/Gridap.jl/pull/577/files#diff-ba506509a89b9ff975762ef98f39e3dd9a7d0ef2b6fdb3b962576fd48abdd161L106

Thus, I guess that I have to also use `cell_orientation` for the evaluation of DoFs to fix this issue (this would match with have it seems they are doing in UFL/Dolphin). However, I still do not understand why. When I do hand-written calculations I do not see why I should change the sign of the DoF. (most probably I am missing something).

fverdugo:
Hi @amartinhuertas !

What is the difference between the `cell_orientation` array you mention and the result of `get_cell_permutations(top::GridTopology,d::Integer)` ? We use the output of the latter in several places to handle non-oriented meshes.

fverdugo:
BTW, a better name for `get_cell_permutations` would be `get_face_permutation` or `get_face_orientation` to be consistent with the names of other routines.

amartinhuertas:
> What is the difference between the cell_orientation array you mention and the result of get_cell_permutations(top::GridTopology,d::Integer) ? We use the output of the latter in several places to handle non-oriented meshes.

I don't know. Can you elaborate a little bit more the output of `get_cell_permutations(top::GridTopology,d::Integer) ` ? 

I guess that you want to know if `cell_orientation` can be computed out of the output of `get_cell_permutations`, is that right? 

`cell_orientation` is a cell-wise array, with as many entries as cells. As far as I understand, this array returns true for a given cell if the geometrical mapping from the reference cell to the physical cells involves a reflection transformation. For example, in 2D,  the y and x axis of the coordinate system of the cell are flipped in physical space in case of a reflection transformation. In such a case, the determinant of the Jacobian is negative. However, in the case of manifolds, the pseudodeterminant does not carry a sign, thus we need another means to determine whether the coordinate system is flipped or not. This is the goal of the `cell_orientation` array. Does this make sense?

fverdugo:
OK, at the end, it is just an array of bools, right? If so, I would name it `cell_is_oriented` (minor thing).

If these "cells" are "facets" of a volume mesh, I guess you can compute the array from the output of `get_cell_permutataions` as
```julia
topo = get_grid_topology(model)
D = num_cell_dims(model)
facet_is_oriented = lazy_map(i->i==1,get_cell_permutations(topo),D-1)
``` 

I.e., perputation index == 1 is for oriented facets.

amartinhuertas:
> OK, at the end, it is just an array of bools, right? If so, I would name it `cell_is_oriented` (minor thing).

Yes. To be more precise, it is a cell-wise array of boolean `Field`s. I tried to have an array of bools instead, but not sure how I can achieve this given the structure of the code.

> If these "cells" are "facets" of a volume mesh, I guess you can compute the array from the output of get_cell_permutataions as

These "cells" are the cells of a volume mesh or surface mesh embedded in a higher dim space, not facets of a volume mesh. In any case, I do not see why what you say is correct, (perhaps because I do not fully know the rationale underlying get_cell_permutations). My concern here is that you may have a value distinct from 1 for cell permutation without having a reflection transformation involved, right?

fverdugo:
Perhaps we can discuss this face-to-face!

amartinhuertas:
>Perhaps we can discuss this face-to-face!

Sure! I guess that this would be very helpful. When are you available? Perhaps @santiagobadia should join as well ... he is aware of this development as well.

amartinhuertas:
The discussion above regarding `cell_orientation` + general manifolds no longer applies. I have undone all changes related to the `cell_orientation` array in this branch. Reason: see commit 276c410ac8ab7e886de7e8e7c38692076db6c174 description. 

This PR depends on #579 and thus must be processed after it.

amartinhuertas:
Hi @fverdugo! 

After PR https://github.com/gridap/Gridap.jl/pull/577, now we can proceed with this PR (i.e., https://github.com/gridap/Gridap.jl/pull/579).

The changes are minor, essentially, replace det by meas, inv by pinvJt, and https://github.com/gridap/Gridap.jl/pull/577/files#diff-ea5ed427dd3dd1fad920c1f2014e8950d5866705bde29f7278c6d93390ae72c1R140. 

These changes are required in order to be able to pass the test that I have introduced here:

https://github.com/gridap/Gridap.jl/pull/577/files#diff-37219e6468e24d7c4b29903834b894aa426a861aa49ef64122068cbd47ceb5daR69

fverdugo (approved):

amartinhuertas:
@santiagobadia ... @fverdugo is happy with the changes introduced in this PR. Are u ok as well? If yes, I think we can accept the PR.

santiagobadia (approved):
All good

PR #579, opened 2021-04-14
Towards a generalized implementation of RT FEs that does not assume an oriented global mesh.

amartinhuertas:
Pull request that stemmed from the discussion in https://github.com/gridap/Gridap.jl/pull/577

Tests are not passing. Spent quite a lot of time debugging the following simple case, and I do not see where is the source of the error. 

```julia
using Test
using Gridap.Geometry
using Gridap.ReferenceFEs
using Gridap.FESpaces
using Gridap.CellData
using Gridap.TensorValues
using Gridap.Fields
using Gridap.Io


order = 0

reffe = ReferenceFE(TRI,raviart_thomas,order)

domain =(0,1,0,1)
partition = (1,1)
model = simplexify(CartesianDiscreteModel(domain,partition))

V = FESpace(model,reffe,conformity=DivConformity())

v(x) = VectorValue(-0.5*x[1]+1.0,-0.5*x[2])
vh = interpolate(v,V)

e = v - vh

Ω = Triangulation(model)
dΩ = Measure(Ω,2*order)

el2 = sqrt(sum( ∫( e⋅e )*dΩ ))
@test el2 < 1.0e-10
```

@santiagobadia ... we should perhaps discuss what I did in this PR so that we can try to find the cause of the error. 

In any case, I still do not understand two things:
 * How we ensure that the outward unit normals in physical space are actually unit normals? I do not see where they are scaled after applying the inverse transpose of the Jacobian to the unit normals in reference space. 
 * Before my mods, the code was also using the signed determinant of the Jacobian for the interior moments. I am not longer using the signed determinant of the Jacobian, but the absolute value. This would affect to order>1 RT FEs.  see https://github.com/gridap/Gridap.jl/pull/579/files#diff-ba506509a89b9ff975762ef98f39e3dd9a7d0ef2b6fdb3b962576fd48abdd161R115


santiagobadia:
Let us talk tomorrow at Monash.

amartinhuertas:
A follow-up on this PR. 

I have simplified the test above, using `v(x) = VectorValue(1,2)` of `v(x) = VectorValue(-0.5*x[1]+1.0,-0.5*x[2])`

When I restrict the the interpolated function to cell number 2, I still get the function with the sign flipped, i.e., [-1 -2].

I have done all computations by hand (see attachment), and double-checked that they match what the program is doing. 

Thus, it might happen which I have missunderstood something in the implementation of RT FEs for arbitrarily oriented meshes

@santiagobadia ... do you know what might be wrong in the notes?

[debugging_rt_fes.pdf](https://github.com/gridap/Gridap.jl/files/6364568/debugging_rt_fes.pdf)

amartinhuertas:
@santiagobadia ... Can it be that the signed determinant of the Jacobian has to be used for the Piola map if and only if you also use it when transforming the differential integral measure? I just thought that it makes sense as the Piola map somehow can be derived from the transformation of domain of integration.  

In any case, not sure about this. If this was true, it would mean that, e.g., fenics  is not using the determinant of the Jacobian when transforming the Domain of integration .... How would you solve the missorientation issue when integrating?
Perhaps with an extra sign flip which comes from info of the mesh?

amartinhuertas:
@fverdugo @santiagobadia ... FYI ...

I followed a different approach for generalizing RT FEs for arbitrarily oriented meshes to the one that I was trying to follow so far. See description of the following commit 3b6b8806a31d6fd0dffba19302db1d4b8701b6da (also added to the source code). 

I think that once the tests pass, the PR is ready to review. 

As a comment, I had to put the following code:

https://github.com/gridap/Gridap.jl/pull/579/files#diff-4a26cbf651e1f78f9b883733fcaf89ad35a3fabaa2e4938c669e5fd333d3ff63R144

 in the `ConformingFESpaces.jl` source file. However, conceptually I guess tha this code should go to the `RaviartThomasRefFEs.jl` file. I could not put it in the former file because of a module cycle (Reffe cannot depend on Model).
Is there any known workaround to this in Julia? Did you have to face similar issues in the past?

amartinhuertas:
>Before my mods, the code was also using the signed determinant of the Jacobian for the interior moments. I am not longer using the signed determinant of the Jacobian, but the absolute value. This would affect to order>1 RT FEs. see https://github.com/gridap/Gridap.jl/pull/579/files#diff-ba506509a89b9ff975762ef98f39e3dd9a7d0ef2b6fdb3b962576fd48abdd161R115

Ok. I think this is no longer a problem. The code was using the signed determinant of the Jacobian both for the Piola Map and the calculation of the interior moments. Now we use the absolute value of the determinant both for the Piola Map and the calculation of interior moments. Thus, the outcome should be same, the tests with high order FEs should pass now.

amartinhuertas:
> I think that once the tests pass, the PR is ready to review.

Ok, the tests passed. PR ready to review.

src/FESpaces/ConformingFESpaces.jl
```diff
@@ -114,15 +116,15 @@ end
...
+  model::DiscreteModel,
+  cell_reffe::AbstractArray{<:ReferenceFE},
+ )
   ctype_reffe, cell_ctype = compress_cell_data(cell_reffe)
   ctype_num_dofs = map(num_dofs,ctype_reffe)
   ctype_ldof_comp = map(reffe->get_dof_to_comp(reffe),ctype_reffe)
   cell_conformity = CellConformity(cell_reffe)
-  cell_shapefuns = lazy_map(get_shapefuns,cell_reffe,cell_map)
-  cell_dof_basis = lazy_map(get_dof_basis,cell_reffe,cell_map)
+  cell_shapefuns = get_shapefuns(model,cell_reffe)
```
fverdugo, line 126: Do you need to pass a full discrete model here? is not enough to do something like 

```
cell_shapefuns = lazy_map(get_shapefuns,cell_reffe,cell_map, cell_flip)
```
?

src/FESpaces/ConformingFESpaces.jl
```diff
@@ -139,6 +141,119 @@ function CellFE(
...
 end
 
+function get_dof_basis(model::DiscreteModel,
+                       cell_reffe::AbstractArray{<:ReferenceFE})
+  lazy_map(get_dof_basis,
+           cell_reffe,
+           get_cell_map(Triangulation(model)))
+end
+
+function get_dof_basis(model::DiscreteModel,
```
fverdugo, line 151: This includes a dependency between discrete model and RaviartThomas reffes.  As a result, RaviartThomas reffes cannot be fully implemented in the ReferenceFEs module, which is sort of confusing. It is also a sign that the current API of ReferenceFEs is not general enough.

Do you think it's possible to fix this? Perhaps it is possible with some refactoring of the ReferenceFEs API.

fverdugo (changes requested):
@amartinhuertas thanks for the PR!

I have some concerns about the introduced coupling between DiscreteModel and Raviart-Thomas reffes. See below.

src/FESpaces/ConformingFESpaces.jl
amartinhuertas: > Do you need to pass a full discrete model here?

We shall agree that there is some work to do at the global FE space level to compute the (cell-wise restriction) of global shape functions and global DoFs in the case of global H(div)-conforming FE spaces (that, e.g., is not required to be done for global grad-conforming FE spaces). In the case of your code snippet, it would be to build the `cell_flip` array out of the model. In the case of the code I have written, those lines of code which I added to `ConformingFESpaces.jl`. The code which is in charge of doing this work cannot go to the `ReferenceFEs` module, otherwise we would have a cycle. I guess that you are trying to avoid this cycle by having interfaces that do not depend on model, and that thus can go to `ReferenceFEs.jl`. But I am not sure if this is a good solution, because of reasons which I try to make clear below. 
 
I understand the concern of having code in `ConformingFESpaces.jl` which depends on the particular kind of global FE space at hand. In order to solve this, I propose to add a new source file, `DivConformingFEspaces.jl`, in FESpaces. See 
https://github.com/gridap/Gridap.jl/pull/579/files#diff-ea5ed427dd3dd1fad920c1f2014e8950d5866705bde29f7278c6d93390ae72c1R1 for a motivation. In a nutshell, this source file implements those customizations of `get_shape_funs` and `get_dof_basis` for the global conforming FE space at hand. 

>  is not enough to do something like 
> ```
> cell_shapefuns = lazy_map(get_shapefuns,cell_reffe,cell_map, cell_flip)
> ```
> ?

I am not sure if this would be convenient. 

First, you would be passing the `cell_flip` array in order to build the (cell-wise restriction) of global shape functions for all global conforming FE spaces, even if it is not needed (e.g., for Grad-Confirming spaces).

Second, it is hard to anticipate if `cell_flip` is the info which will be needed in order to compute the (cell-wise restriction) of global shape functions for those global conforming FE spaces which we may want to implement  in the future. 

By having a `get_shapefuns` interface with model, and the ReferenceFE, we have more flexibility in defining the global shape functions according to the global FE space at hand. If the name is confusing, we can perhaps think of another name, instead of reusing the one that is used in order to compute the reference FE shape functions. 

Let me know what do you think.

src/FESpaces/ConformingFESpaces.jl
amartinhuertas: See https://github.com/gridap/Gridap.jl/pull/579#discussion_r621828947

santiagobadia: I have been talking to @amartinhuertas this morning, and I think this is the way to go. Div-conforming FE spaces require to modify the reference basis when transformed to the local FE basis. That is the identity for Lagrangian FEM (grad-conforming) but not the case (in general) for other FEs like div or curl-conforming ones.

This is something that cannot be at the ref FE level (it is a global thing, to enforce div-conformity in the global space).

I also don't like to have dispatching wrt RT in the `ConformingFESpace` file. We have also learned that the __only__ modification that is required to go to div-conforming are these two functions that compute the global DOF and shape functions on physical cells. So, it seems to me that the most neat solution is to create a file for these FE spaces that simply implement these functions.

It is in the same spirit as e.g. discontinuous Galerkin and conforming FE spaces, which fill call the unconstrained FE constructor in different ways. We could also do the dispatching at that level for div-conforming, but since the code is the same with only the previous exception, I would just implement the functions that need dispatching.

Cheers,

--Santi

fverdugo: We can deal with this If it is necessary and I would also reame the versions taking the discrete model to `get_cell_shapefuns` and `get_cell_dof_basis` like other functions returning cell-wise values.

Back in the day, `get_shapefuns` and `get_dof_basis` had only a single argument (the reference FE), but we included a second argument (the geometrical map) to handle RT and Nedelec. Now these FE types will be handled differently, so perhaps we can revert to the original definition of `get_shapefuns` and `get_dof_basis` with just a single argument.  This would simplify the implementation of Lagrangian reffes, which became a bit dirty after introducing RT and Nedelec.

In summary, these would the API

```julia
# Abstract methods in ReferenceFEs module
get_shapefuns(reffe::ReferenceFE)
get_dof_basis(reffe::ReferenceFE)

# Methods with a default implementation in terms of previous ones (which will work for e.g. Lagrangian),
# but they can be overloaded by new ReferenceFE types
get_cell_shapefuns(model::DiscreteModel,cell_reffe::AbstractArray{<:ReferenceFE}))
get_cell_dof_basis(model::DiscreteModel,cell_reffe::AbstractArray{<:ReferenceFE}))

# Methods to be added for RT FEs

# 1) Helper methods specific for RT
get_shapefuns(reffe::GenericRefFE{RaviartThomas},map::Field,signflip::Integer)
get_dof_basis(reffe::GenericRefFE{RaviartThomas},map::Field,signflip::Integer)

# 2) Specialization of Abstract methods in ReferenceFEs module
get_shapefuns(reffe::GenericRefFE{RaviartThomas}) = get_shapefuns(reffe,GenericField(identity),1)
get_dof_basis(reffe::GenericRefFE{RaviartThomas}) = get_dof_basis(reffe,GenericField(identity),1)

# 3) Specialization of cell-wise (like already implemented in the PR)
get_cell_shapefuns(model::DiscreteModel,cell_reffe::AbstractArray{<:GenericRefFE{RaviartThomas}}))
get_cell_dof_basis(model::DiscreteModel,cell_reffe::AbstractArray{<:GenericRefFE{RaviartThomas}}))
```

What do you think?

src/FESpaces/DivConformingFESpaces.jl
```diff
@@ -0,0 +1,103 @@
...
+    sign_flip = get_sign_flip(model, cell_reffe)
+    lazy_map(get_shapefuns,
+             cell_reffe,
+             get_cell_map(Triangulation(model)),
+             lazy_map(Broadcasting(ConstantField), sign_flip))
+end
+
+function get_sign_flip(model::DiscreteModel,
+                       cell_reffe::AbstractArray{<:GenericRefFE{RaviartThomas}})
+    lazy_map(get_sign_flip,
```
fverdugo, line 28: (Minor comment). Perhaps, it is easier to pre compute the the cell-wise sign flip into a standard julia vector of `Int8` instead to try to make it lazy. It would be also significantly faster if you need to perform several loops over the vector, e.g. in non-linear problems

src/FESpaces/ConformingFESpaces.jl
santiagobadia: Agreed

santiagobadia: Btw, the sign is not +1 or -1, not an `Int` but an array with as many +1/-1 as DOFs in the RefFE

src/FESpaces/DivConformingFESpaces.jl
santiagobadia: The array is not as simple, it is an array of arrays of `Int`. So, I would not proceed this way.

src/FESpaces/ConformingFESpaces.jl
fverdugo: Ok! I have updated the code snipped

src/FESpaces/DivConformingFESpaces.jl
fverdugo: Then better lazy!

I would slightly refactor the definition of the kernel from which the lazy array is defined to match other similar kernels in Gridap. Here, the cell operation depends on a global state (the model). In this case, we usualy include the global state in a struct:

```julia
struct SignFlipMap{T} <: Map
  model::T
end

function return_cache(k::SignFlipMap,reffe,cell)
  model = k.model
 # same code as below
end

function evaluate!(cache,k::SignFlipMap,reffe,cell)
 # same code as below
end

# usage
lazy_map(SignFlipMap(model),cell_reffe,cell_ids)

```

src/FESpaces/DivConformingFESpaces.jl
```diff
@@ -0,0 +1,103 @@
...
+                  cell_id::Integer)
+
+    cell_wise_facets_ids,
+    cache_cell_wise_facets_ids,
+    cells_around_facets,
+    cache_cells_around_facets,
+    sign_flip = cache
+
+    setsize!(sign_flip, (num_dofs(reffe),))
+    sign_flip .= false
```
fverdugo, line 82: Replace this:
```julia
    sign_flip = cache
    setsize!(sign_flip, (num_dofs(reffe),))
```
by
```julia
    setsize!(cache, (num_dofs(reffe),))
    sign_flip = cache.array
```
Then `sign_flip` (and the return of this function) will be a intrinsic julia array. This is also a common pattern used in Gridap to hide the cached vector from the user of this function.

src/FESpaces/ConformingFESpaces.jl
amartinhuertas: I used `Bool`, not `Int`. (true-> -1 false -> 1; not a big deal)

src/FESpaces/DivConformingFESpaces.jl
amartinhuertas: Done in 1685e4d99808277dba2e2f7062deb14e6f969bdb

src/FESpaces/DivConformingFESpaces.jl
amartinhuertas: Done in 1685e4d99808277dba2e2f7062deb14e6f969bdb

src/FESpaces/ConformingFESpaces.jl
amartinhuertas: Done in 1685e4d99808277dba2e2f7062deb14e6f969bdb

fverdugo:
@amartinhuertas the renaming `get_cell_dof_bais` -> `get_fe_basis` is not correct.

It should be something like
-  `get_cell_shapefuns(::FESpace)` -> `get_fe_basis`
-  `get_cell_shapefuns_trial(::FESpace)` -> `get_trial_fe_basis`
- `get_cell_dof_basis(::FESpace)` ->  `get_fe_dof_basis`

There are the current functions in FESpace that are called `get_cell_x` and do not return an array

fverdugo:
After the renaming, you can introduce the new functions `get_cell_shapefuns(model,cell_reffes)` and `get_cell_dof_basis(model,cell_reffes)`.

So the final API would be:
```julia
get_fe_basis(fe_space) -> FEBasis
get_trial_fe_basis(fe_space) -> FEBasis
get_fe_dof_basis(fe_space) -> CellDof
get_cell_shapefuns(model,cell_reffes) -> AbstractArray{<:AbstractVector{<:Field}}
get_cell_dof_basis(model,cell_reffes) -> AbstractArray{<:AbstractVector{<:Dof}}
```

amartinhuertas:
> @amartinhuertas the renaming get_cell_dof_bais -> get_fe_basis is not correct.

Ooops! You are right (too much rushing). Now it should be solved. 

Note also the following change that I had to do in order to pass the tests:

https://github.com/gridap/Gridap.jl/pull/579/commits/45d5f60e76392c0600e4147ef319e81c83a706b0

I am not sure if it still makes sense to have `trian` as a parameter in the interface of `compute_cell_space`. What do you think? should we remove this parameter ?

amartinhuertas:
Should we add deprecated warnings/errors for the renamed functions?

src/FESpaces/ConformingFESpaces.jl
```diff
@@ -189,10 +199,8 @@ end
 
 function _compute_cell_space(
```
fverdugo, line 200: Is this method required after the refactoring? I would remove it since it only works for lagrangian FEs and the method `_compute_cell_space` defined for `CellFE` should be enough.

fverdugo: That is

- Keep this: ` _compute_cell_space(cell_fe::CellFE)` (removing trian)
- Delete this: ` _compute_cell_space(cell_reffe::AbstractArray{<:ReferenceFE}, trian::Triangulation)
`

fverdugo:
> I am not sure if it still makes sense to have trian as a parameter in the interface of compute_cell_space. What do you think? should we remove this parameter ?

It looks like you still need `trian` to build the `SinglefieldFEFunction`

fverdugo:
> Should we add deprecated warnings/errors for the renamed functions?

It would be very convenient! I would use informative errors more than deprecation warnings.

fverdugo:
@amartinhuertas I have merged a major refactoring into master (PR #583). It has introduced some conflicts with this branch and I have fixed it. Hope the tests pass again!

amartinhuertas:
> @amartinhuertas I have merged a major refactoring into master (PR #583). It has introduced some conflicts with this branch and I have fixed it. Hope the tests pass again!

Thanks! They don't, but I will fix it. I am also doing other changes.

fverdugo:
Please pull from master again. I have updated the NEWS.md.

src/FESpaces/ConformingFESpaces.jl
amartinhuertas: > Is this method required after the refactoring? 

I think so, it is called at several places. The method returns a pair with the `fe_basis` and the `dof_basis` (CellDatum objects). For conforming FE spaces, these objects are extracted out of `CellFE`. For other FE spaces (I would say made invariably of Lagrangian FEs), these objects are built out of an array of RefFEs, and the triangulation. 

Thus: (1) the method cannot be removed. (2) the trian parameter cannot be removed. (3) the method does not only work with lagrangian FEs.  

On the other hand, the `trian` parameter can be removed for both variants of `_compute_cell_space` (I would keep the second variant, it is still helpful).

fverdugo: There is something that I don't understand. After the refactoring, you need a model to get the cell-wise shape functions from the reference FEs in general. Thus, you cannot exteact valid cell-wise shape functions from the arguments in this method `_compute_cell_space(cell_reffe::AbstractArray{<:ReferenceFE}, trian::Triangulation)` in the general case... the model is missing.

fverdugo: is perhaps being used in discontinus spaces? Even in discontinus spaces, I think you need to apply the piola transformation (perhaps the sign flip is not hat crucial in this case). Thus, you still need the model.

We need to remove the method _compute_cell_space(cell_reffe::AbstractArray{<:ReferenceFE}, trian::Triangulation) since i believe it does not make sense anymore after the refactoring. Just go to all the calles of it (or the calles of compute_cell_space) and transform the `AbstractArray{<:ReferenceFE}` into a `CellFE` using a model (which should be over there). Then the method could be removed.

amartinhuertas: > There is something that I don't understand. After the refactoring, you need a model to get the cell-wise shape functions from the reference FEs in general. Thus, you cannot extract valid cell-wise shape functions from the arguments in this method _compute_cell_space(cell_reffe::AbstractArray{<:ReferenceFE}, trian::Triangulation) in the general case... the model is missing.

`compute_cell_space` (that in turn calls `_compute_cell_space`) is also called from FESpaces which are not created out of a model (e.g., DG FE Spaces), but out of a triangulation instead. In this case, compute_cell_space gets the array of RefFEs and the triangulation. (i.e., you cannot pass the model, because it is not available).

amartinhuertas: In the case of FESpaces which are created out of a model, `compute_cell_space` gets a `CellFE` instance (instead of an array of reffes, that in turn has inside as member variables the shape functions and the dof basis created out of a model.

fverdugo: It looks like you can only create a dg space from the triangulation for Lagranges FEs, not for Raviart-Thomas or in general.

After the refactoring, the construction of the cell-wise shape functions and dofs basis is coupled with the discrete model, thus it is not possible to build a dg space from a triangulaiton in general now....

Another reason why i don't like this new coupling, but It seems that we need to deal with it.

fverdugo: >  is also called from FESpaces which are not created out of a model (e.g., DG FE Spaces)

It looks like we need to remove the constructor of the DG space from a `Triangulation`, and replace it from 2 new constructors, one that takes a `DiscreteModel`, and another that takes the a `CellFE`. The former will call the latter.

amartinhuertas: > We need to remove the method _compute_cell_space(cell_reffe::AbstractArray{<:ReferenceFE}, trian::Triangulation) since i believe it does not make sense anymore after the refactoring. Just go to all the callers of it (or the calles of compute_cell_space) and transform the AbstractArray{<:ReferenceFE} into a CellFE using a model (which should be over there). Then the method could be removed.

I guess that this comment no longer applies, right?

amartinhuertas: > You can only create a dg space from the triangulation for Lagranges FEs, not for Raviart-Thomas or in general.

> After the refactoring, the construction of the cell-wise shape functions and dofs basis is coupled with the discrete model, thus it is not possible to build a dg space from a triangulaiton in general now....

Could this be addressed defining variants of `get_cell_shapefuns` and `get_cell_dof_basis` for an array of reffes + triangulation?

fverdugo: It still applies. the DG space cannot constructed from a triangulation anymore.

We need to remove the constructor of the DG space from a Triangulation, and replace it from 2 new constructors, one that takes a DiscreteModel, and another that takes the arrays cell_shapefuns, cell_dof_basis. The former will call the latter.

In addition, remove this 
`_compute_cell_space(cell_reffe::AbstractArray{<:ReferenceFE}, trian::Triangulation)`
and replace it by 
`_compute_cell_space(cell_shapefuns,cell_dof_basis,trian)` (trian only used to build the CellDatum objects)

The same pattern should be used in other FEspaces that are build from a triangulation (except the spaces that are tailored for Lagrangian elements)

fverdugo: > Could this be addressed defining variants of get_cell_shapefuns and get_cell_dof_basis for an array of reffes + triangulation?

I would not do this for the same reason now we only have get_shapefuns(reffe) and not `get_shapefuns(reffe,map)`

amartinhuertas: > It looks like we need to remove the constructor of the DG space from a Triangulation, and replace it from 2 new constructors, one that takes a DiscreteModel, and another that takes the arrays cell_shapefuns, cell_dof_basis. The former will call the latter.

For my understanding. Why the previous code was trying to stick into FE spaces built out of Triangulations as far as possible? What do you gain with that? (I am sure there is something, but I do not see it now). I guess that you gain generality/flexibility ... but do you have use cases in mind where this is necessary/relevant?

fverdugo: > Why the previous code was trying to stick into FE spaces built out of Triangulations as far as possible?

Because a triangulation is cheaper tho build than a discrete model. But it practice we always build a discrete model, so not bit deal...

I am even considering to add a discrete model into CellDatum objects instead of a triangulation (it would simplify the implementation of mixed dimensional pdes, and perhaps hybrid methods also)

amartinhuertas: Ok. So recap after this long discussion, it seems we have came up with two options:

1. Discontinuous FE spaces (except those built out of Lagrangian RefFEs) must be built out of a model.
  * Pros: we stick into a single interface for cell_shapefuns, cell_dof_basis.
  * Cons: forced to use a model (do we loose something?). We end up doing work which is not needed, e.g., sign flip. 
2. Define variants of get_cell_shapefuns and get_cell_dof_basis for an array of reffes + triangulation.
* Pros: we preserve the invariant of being able to set up the FE Space out of triangulation. We can skip doing work which is not needed for DG spaces (e.g., change of basis for Lagrangian FEs, sign flip for RT FEs). 
* Cons: "I would not do this for the same reason now we only have get_shapefuns(reffe) and not get_shapefuns(reffe,map)" the dispatching among Discontinuous and Continuous FE Spaces is based on the type of geometry object (i.e., model versus triangulation). This seems to me a bit ugly.  

Wrote for documentation purposes. If you feel that 1. is the way to go, we can proceed. Perhaps we can think a little bit more. I am going to stop working now on this. (it's late already :-))

amartinhuertas:
> Thanks! They don't, but I will fix it. I am also doing other changes.

FYI ... tests are now passing.

amartinhuertas:
> It would be very convenient! I would use informative errors more than deprecation warnings.

Done in 3d36a5731dd46e2cf21637394d54a2fd951af407

fverdugo:
> Ok. So recap after this long discussion, it seems we have came up with two options:
Discontinuous FE spaces (except those built out of Lagrangian RefFEs) must be built out of a model.
Pros: we stick into a single interface for cell_shapefuns, cell_dof_basis.
Cons: forced to use a model (do we loose something?). We end up doing work which is not needed, e.g., sign flip.
Define variants of get_cell_shapefuns and get_cell_dof_basis for an array of reffes + triangulation.
Pros: we preserve the invariant of being able to set up the FE Space out of triangulation. We can skip doing work which is not needed for DG spaces (e.g., change of basis for Lagrangian FEs, sign flip for RT FEs).
Cons: "I would not do this for the same reason now we only have get_shapefuns(reffe) and not get_shapefuns(reffe,map)" the dispatching among Discontinuous and Continuous FE Spaces is based on the type of geometry object (i.e., model versus triangulation). This seems to me a bit ugly.

Summarizing:

- These are the signatures we need.
```julia
get_shapefuns(reffe::ReferenceFE)
get_dof_basis(reffe::ReferenceFE)
get_cell_shapefuns(model::DiscreteModel,cell_reffe::AbstractArray{<:ReferenceFE})
get_cell_dof_basis(model,cell_reffe::AbstractArray{<:ReferenceFE})
compute_cell_space(cell_shapefuns::AbstractArray{<:Field},cell_dof_basis::AbstractArray{<:Dof},trian::Triangulation)
compute_cell_space(cellfe::CellFE,trian::Triangulation)=compute_cell_space(cellfe.cell_shapefuns,cellfe.cell_dof_basis,trian)
```

-  All signatures using a `Triangulation` to generate the cell-wise shape funs and dof basis, should be removed, since now we need a `DiscreteModel`.

- The current DG space constructor
```julia
function _DiscontinuousFESpace(
  vector_type::Type,
  trian::Triangulation,
  cell_fe::CellFE)
```
is already OK since it takes a `CellFE`. **Do not replace trian for model**. The model was already used to generate the 
 cell-wise shape funs and dof basis stored in the `CellFE`.

amartinhuertas:
> These are the signatures we need.
> get_shapefuns(reffe::ReferenceFE)
> get_dof_basis(reffe::ReferenceFE)
> get_cell_shapefuns(model::DiscreteModel,cell_reffe::AbstractArray{<:ReferenceFE})
> get_cell_dof_basis(model,cell_reffe::AbstractArray{<:ReferenceFE})
> compute_cell_space(cell_shapefuns::AbstractArray{<:Field},cell_dof_basis::AbstractArray{<:Dof},trian::Triangulation)
> compute_cell_space(cellfe::CellFE,trian::Triangulation)=compute_cell_space(cellfe.cell_shapefuns,cellfe.cell_dof_basis,trian)

> All signatures using a Triangulation to generate the cell-wise shape funs and dof basis, should be removed, since now we need a DiscreteModel.

As far as I could detect looking at the code, the only call which is currently (indirectly) using a Triangulation to generate the cell-wise shape funs and dof basis is the one in this line

https://github.com/gridap/Gridap.jl/blob/3d36a5731dd46e2cf21637394d54a2fd951af407/src/FESpaces/CLagrangianFESpaces.jl#L112

BTW, which is the purpose of `CLagrangianFESpaces` ? (I have the feeling that they cannot be built using the general `FESpace` constructor.) They are not built out of a model, but using a grid instead. Thus, for these spaces, we cannot use the `get_cell_shapefuns` and `get_cell_dof_basis`. How do we bypass this? 

> The current DG space constructor

> function _DiscontinuousFESpace(
>   vector_type::Type,
>  trian::Triangulation,
>   cell_fe::CellFE)
> is already OK since it takes a CellFE. Do not replace trian for model. The model was already used to generate the
> cell-wise shape funs and dof basis stored in the CellFE.

Ok, I see now looking very carefully at the code. Thus, we were somehow confused (I am not sure what caused the confusion). DiscontinuousFESpaces are also built out of a model, thus we can stick into a single interface for `get_cell_shapefuns` and `get_cell_dof_basis`. The only drawback that I see (if am not wrong), is that we end up doing the same work for Conforming and NonConforming spaces with regards to the setup of the cell shape funs and cell Dof Basis. I am not sure which are the consequences of this in the most general case, and thus whether we can live with that or not.

amartinhuertas:
> The only drawback that I see (if am not wrong), is that we end up doing the same work for Conforming and NonConforming spaces with regards to the setup of the cell shape funs and cell Dof Basis. I am not sure which are the consequences of this in the most general case, and thus whether we can live with that or not.

@fverdugo ... FYI ... I have been talking a little while with @santiagobadia in regards to this. We are positive that this can be considered adding a third parameter to `CellFE(model,reffes,conformity)` and `get_cell_shapefuns` and `get_cell_dof_basis` with the notion of conformity of the global FE space at hand (i.e., `:L2`, `:H1`, `:Hcurl`, etc.).

On the other hand, the way the `conformity` parameter is handled from the root `FESpace` constructor, top-bottom, seems quite intrincate/hard-to-read (`nothing`, `:default`, etc.). Why all this is necessary? (perhaps we are missing something).
How we may simplify it? (e.g., can we set the `conformity` ASAP, instead of propagating `nothing` and `:default` down the call tree? ) I am going to open a separate issue on Gripad so that we can discuss this separately from this work. 

> Thus, for these spaces, we cannot use the get_cell_shapefuns and get_cell_dof_basis. How do we bypass this?

I think I will just call `lazy_map(get_shapefuns, cell_reffes)` and `lazy_map(get_dof_basis, cell_reffes)` right before calling `compute_cell_space`.

fverdugo:
> @fverdugo ... FYI ... I have been talking a little while with @santiagobadia in regards to this. We are positive that this can be considered adding a third parameter to CellFE(model,reffes,conformity) and get_cell_shapefuns and get_cell_dof_basis with the notion of conformity of the global FE space at hand (i.e., :L2, :H1, :Hcurl, etc.).

OK. this would allow us to perform some optimizations e.g. for RT and L2. Use the types describing the conformity (e.g. `L2Conformity()` ) instead of symbols (otherwise you cannot do dispatch).

> I think I will just call lazy_map(get_shapefuns, cell_reffes) and lazy_map(get_dof_basis, cell_reffes) right before calling compute_cell_space.

Yes, when it makes sense (e.g. in Lagrangian spaces).

> BTW, which is the purpose of CLagrangianFESpaces ?

It implements the particular case when you want a continuous Lagrangian space of the same order of the grid. It is much more efficient than the general conforming FE space since one can re-use the cell connectivity as the cell dofs. Also provides the maps between nodes and dofs (not given by the general conforming FE space) which are needed in some contexts. (Only implemented now for no Dirichlet conditions)

fverdugo:
> On the other hand, the way the conformity parameter is handled from the root FESpace constructor, top-bottom, seems quite intrincate/hard-to-read (nothing, :default, etc.). Why all this is necessary? (perhaps we are missing something).

Change this signature

```julia
function FESpace(
  model::DiscreteModel,
  cell_fe::CellFE;
  conformity=nothing,
  labels = get_face_labeling(model),
  dirichlet_tags=Int[],
  dirichlet_masks=nothing,
  constraint=nothing,
  vector_type::Union{Nothing,Type}=nothing)
```
to 
```julia
function FESpace(
  model::DiscreteModel,
  cell_fe::CellFE,
  cell_conf::CellConformity;
  labels = get_face_labeling(model),
  dirichlet_tags=Int[],
  dirichlet_masks=nothing,
  constraint=nothing,
  vector_type::Union{Nothing,Type}=nothing)
```
(conformity is now a positional argument of type CellConformity )

Then, change this
```julia
function FESpace(
  model::DiscreteModel,
  cell_reffe::AbstractArray{<:ReferenceFE};
  conformity=nothing,
  kwargs...)
  if conformity in (nothing, :default, :L2, L2Conformity())
    conf = conformity
  else
    conf = CellConformity(cell_reffe,conformity)
  end
  cell_fe = CellFE(model,cell_reffe)
  FESpace(model,cell_fe;conformity=conf,kwargs...)
end
```
to
```julia
function FESpace(
  model::DiscreteModel,
  cell_reffe::AbstractArray{<:ReferenceFE};
  conformity=nothing,
  kwargs...)
  trian = Triangulation(model)
  cell_map = get_cell_map(trian)
  reffe = testitem(cell_reffe)
  c = Conformity(reffe,conformity)
  cell_conf = CellConformity(cell_reffe,c)
  cell_fe = CellFE(cell_map,cell_reffe,c)
  FESpace(model,cell_fe,cell_conf;kwargs...)
end
```
(here i am assuming that `CellFE`already takes a `Conformity` object in the last argument)

This should simplify things.

amartinhuertas:
> This should simplify things.

Ok, perfect. Thanks for the proposal. I can do it. Would you make this change here or in a separate PR?

fverdugo:
> Ok, perfect. Thanks for the proposal. I can do it. Would you make this change here or in a separate PR?

Here please! Update also this method accordingly https://github.com/gridap/Gridap.jl/blob/b0ddbbf2c5da9c50ca065764ae4b86647569ee36/src/FESpaces/FESpaceFactories.jl#L65

On the other hand, are these signatures implemented?
```julia
CellFE(
  model::DiscreteModel,
  cell_reffe::AbstractArray{<:ReferenceFEs},
  conf::ConformityStyle)

get_cell_shapefuns(
   model::DiscreteModel,
   cell_reffe::AbstractArray{<:ReferenceFEs},
   conf::ConformityStyle)

get_cell_dof_basis(
   model::DiscreteModel,
   cell_reffe::AbstractArray{<:ReferenceFEs},
   conf::ConformityStyle)
``` 
 ?

amartinhuertas:
> Update also this method accordingly

Not sure I undertand. This method has an optional keyword argument called `constraint`. `constraint` and `conformity` are different concepts, right?

> On the other hand, are these signatures implemented?

Not yet. I am working with other things. But they will be soon.

amartinhuertas:
Hi @fverdugo ... I have taken a more closer look to what you proposed in regards to the signature changes of the `FESpace` factory method. 

I have detected/foresee two issues/knots still to be tied:
  1. How are we going to implement this line (https://github.com/gridap/Gridap.jl/blob/3d36a5731dd46e2cf21637394d54a2fd951af407/src/FESpaces/FESpaceFactories.jl#L30) if `conformity` is of type `CellConformity` ?
  2.  I guess that the interface of `FiniteElements` should be modified such that it also gets `conformity=nothing` as parameter.

I guess that 2. is easy to fix. But what about 2., it seems not hat easy right?

amartinhuertas:
> Not sure I understand. This method has an optional keyword argument called constraint. constraint and conformity are different concepts, right?

Ok. Now I understand. `conformity` was an optional keyword argument. With your proposal it becomes a positional argument, and thus I have to add it as well to the left of the ; in the function that you pointed out.

amartinhuertas:
>   1. How are we going to implement this line (https://github.com/gridap/Gridap.jl/blob/3d36a5731dd46e2cf21637394d54a2fd951af407/src/FESpaces/FESpaceFactories.jl#L30) if `conformity` is of type `CellConformity` ?

@fverdugo ... I have been thinking into this a little while, while performing a more deep immersion into the rationale underlying the `conformity` parameter of the FESpace constructor, the `Conformity` and `CellConformity` types, and their interplay with ReferenceFE instances.
I think now I understand the big picture. Even a `ReferenceFE` handles a notion of conformity internally, which is defined when the REferenceFE instance is built, you may override this when building the `FESpace` through the `conformity` parameter. The `CellConformity` data type stores, separately from the `ReferenceFE`, the dof ownership info corresponding to the notion of conformity specified via the `conformity` parameter (actually for each cell topology, and reference element instance, if several).

Thus, I foresee two possible solutions to fix the issue that I quoted at the beginning of this post:
 1. To have type `Conformity` as a trait of `CellConformity`, and do type distpaching via the `FESpace` constructor (to select among conforming versus discontinuous FE spaces).
 2. To have an instance of type `Conformity` as a member variable of `CellConformity`, and have an if statament as we have now. 

In think I will go for 2., instead of 1., but I prefer that you let me know your thoughs, just to double check that I am not missing anything here. 

Thanks!

amartinhuertas:
> In think I will go for 2., instead of 1., but I prefer that you let me know your thoughs, just to double check that I am not missing anything here.

I already implemented and pushed 2., and the full set of changes that we discussed along the way. Code ready for review (up to tests, which are still running).

Some observations with regards to the code in its current state (in order to facilitate code review):

* `CellConformity` has a new member variable of type `Conformity`.
*  The `CellFE(model,cell_reffes,cell_conformity)` function receives a `CellConformity` instance (instead of a `Conformity` instance as discussed above). It did not make sense to me to recreate the `CellConformity` instance inside the `CellFE` function if it was created right before calling `CellFE` (perhaps I am missing a use case that you foresaw this). 
* The `FESpace` constructor does not have a `cell_conformity` parameter (as discussed above). I did not see why this is necessary  if `CellFE` already has `cell_conformity` as a member variable. (perhaps I am missing a use case that you foresaw this). 
* I added an optional keyword argument `conformity` to the `FiniteElements` constructor. 
* Several methods of `CellConformity` are no longer necessary (currently commented out just in case we have to roll back some of these changes).
* In `PhysicalFEsTests.jl`, as a consequence of the changes introduced, I was forced to create a different `CellFE` instance for `L2` and nothing (i.e., `H1`) conformity.  Before the changes in this PR, the same `CellFE` instance was re-used for both calls to `FESpace`. I guess that know this makes sense because you may have different `CellFE` instances depending on the global notion of conformity of the FE space.

amartinhuertas:
>I already implemented and pushed 2., and the full set of changes that we discussed along the way. Code ready for review (up to tests, which are still running).

Test now passing!

fverdugo:
Hi @amartinhuertas!

The purpose of `CellFE` and `CellConformity` is to store the raw data from which a general conforming space is constructed. These are potentially more general than a cell wise vector of reference fes plus a `Comformity` object (i.e., the user could define a strange spaces that is continuous only at certain facets in a way that cannot be defined in terms of reference fes). Thus, any metadata like a `Conformity` object should not be included in `CellConformity`.

Why do you need to add `Comformity` inside `CellConformity`?

amartinhuertas:
> The purpose of CellFE and CellConformity is to store the raw data from which a general conforming space is constructed. These are potentially more general than a cell wise vector of reference fes plus a Comformity object (i.e., the user could define a strange space that is continuous only at certain facets in a way that cannot be defined in terms of reference fes). Thus, any metadata like a Conformity object should not be included in CellConformity.

Ok, thanks for the clarification. I think now I understand what you say. Thus, as I suspected (but didn't see), you were foreseeing uses cases that I am preventing with the last changes that I introduced (and should thus rollback).

 > Why do you need to add Comformity inside CellConformity?

Point 1. i in https://github.com/gridap/Gridap.jl/pull/579#issuecomment-830057562. See also https://github.com/gridap/Gridap.jl/pull/579#issuecomment-831004875

fverdugo:
> How are we going to implement this line if conformity is of type CellConformity ?

I would provide 2 constructors:
- one taking `CellFE` and `CellConformity` (the low level one)
- another taking a vector of `ReferenceFEs` and `Conformity`

 (in the 2 cases the second object can be optional since a default value can be taken from the first one)
 
 If the user wants to construct a FE space from a vector of `ReferenceFEs` plus a `CellConformity`, one can always build a `CellFE` from the reference fes and use the 1st constructor.

fverdugo:
> If the user wants to construct a FE space from a vector of ReferenceFEs plus a CellConformity, one can always build a CellFE from the reference fes and use the 1st constructor.

One can even add these code into a 3rd constructor if necessary.

amartinhuertas:
> I would provide 2 constructors:

Ok, I see. I will come back to you soon with (hopefully) the final version.

amartinhuertas:
> Ok, I see. I will come back to you soon with (hopefully) the final version.

Ok, done. Remarks:

* `CellConformity` no longer has a member variable of type `Conformity`.
* We have now a `CellFE(model,cell_reffes,conformity=nothing)` function (as we sketched at the beginning, except for the fact that conformity is though to be either nothing or a valid symbol, instead of an instance of type `Conformity`).
* *"The FESpace constructor does not have a cell_conformity parameter (as discussed above). I did not see why this is necessary if CellFE already has cell_conformity as a member variable. (perhaps I am missing a use case when you foresaw this)."* This no longer applies. The FESpace constructor now has an optional cell_conformity argument.
* *"I added an optional keyword argument conformity to the FiniteElements constructor."* This still applies.
* *"Several methods of CellConformity are no longer necessary (currently commented out just in case we have to roll back some of these changes)."* This still applies. Not sure why. Can we remove the methods of `CellConformity` which we are not calling?
* *"In PhysicalFEsTests.jl, as a consequence of the changes introduced, I was forced to create a different CellFE instance for L2 and nothing (i.e., H1) conformity. Before the changes in this PR, the same CellFE instance was re-used for both calls to FESpace. I guess that know this makes sense because you may have different CellFE instances depending on the global notion of conformity of the FE space."* This still applies.

fverdugo:
> except for the fact that conformity is though to be either nothing or a valid symbol

The input should be a `Conformity` object, which is the extensible way of doing dispatch. Why you need a symbol?

amartinhuertas:
> The input should be a Conformity object, which is the extensible way of doing dispatch. Why you need a symbol?

It is not actually needed. I can (and I will) definitely use a Conformity instance as a third parameter of the CellFE constructor.  I did not do it because of two reasons:
* I waived the construction of the Conformity object to the caller, but construct it inside of the CellFE constructor instead. 
* Dispatching on Conformity is already done inside the CellFE constructor, at the get_cell_dof_basis and get_cell_shapefuns level, and I did not foresee a use case in which the gained generality by using a Conformity parameter at the CellFE constructor would actually be used.

In any case, I agree that using a Conformity parameter opens the door for (unforeseen)  possibilities in the future, so that I will follow your advice.

amartinhuertas:
>  I can (and I will) definitely use a Conformity instance as a third parameter of the CellFE constructor. 

Done in  5e1afaf2a590471adf392554cd8a8fb6c6942d86. 

Please continue your code review taking https://github.com/gridap/Gridap.jl/pull/579#issuecomment-831844412 as a basis. You will see there some questions that require an answer.

src/FESpaces/FESpaceFactories.jl
```diff
@@ -1,8 +1,8 @@
 
 function FESpace(
   model::DiscreteModel,
-  cell_fe::CellFE;
-  conformity=nothing,
+  cell_fe::CellFE,
+  cell_conformity::CellConformity=CellConformity(cell_fe);
```
santiagobadia, line 5: I would not include a positional argument here. `CellConformity` is in `CellFE` and I don't think we want/consider the case in which we have different `CellConformity` in `CellFE` than the one passed in this constructor.

santiagobadia (changes requested):

src/FESpaces/FESpaceFactories.jl
amartinhuertas: ok, agreed. I do not see either why this is necessary, and can be even contraproductive.

amartinhuertas:
Ok, done. I think the PR is more or less ready to be merged up to following comment that still needs to be addressed:

> "Several methods of CellConformity are no longer necessary (currently commented out just in case we have to roll back some of these changes)." This still applies. Not sure why. Can we remove the methods of CellConformity which we are not calling?

what do u think @fverdugo?

amartinhuertas:
I am in particular referring to these three methods:

```
function CellConformity(cell_fe::CellFE,cell_conf::Nothing)
  cell_fe.cell_conformity
end

function CellConformity(cell_fe::CellFE,cell_conf::CellConformity)
  @assert length(cell_fe.cell_ctype) == length(cell_conf.cell_ctype)
  cell_conf
end

# function CellConformity(cell_fe::CellFE,conformity::Union{Symbol,Conformity})
#   if conformity in (L2Conformity(),:L2)
#     @notimplemented
#   elseif conformity == :default
#     cell_fe.cell_conformity
#   else
#     @unreachable """\n
#     Argument conformity should be either :default, :L2, or L2Conformity()
#     """
#   end
# end
```

fverdugo:
> I am in particular referring to these three methods:

I agree that they are not longer needed. They can be removed.

src/FESpaces/DivConformingFESpaces.jl
```diff
@@ -0,0 +1,171 @@
...
+# * When we interpolate a function into the global FE space, and we perform the cell-wise
+#   DoF values to global DoF values gather operation, we can either extract the global DoF value
+#   from the master or slave cell without worrying about the sign.
+# * When we evaluate a global FE function, and we perform the global DoF values to
+#   cell-wise DoF values scatter operation, we don't have to worry about the sign either.
+#   On the slave cell, we will have both the sign of the DoF value, and the sign of the
+#   shape function corresponding to the global DoF.
+# * We do NOT have to use the signed determinant, but its absolute value, in the Piola Map.
+
+function get_cell_dof_basis(model::DiscreteModel,
```
fverdugo, line 27: It looks that when using RT FEs with L2 conformity, the default cell shape funs will be generated (i.e. the ones without  the transformation).   Is this enough?

src/Geometry/AppendedTriangulations.jl
```diff
@@ -70,9 +70,9 @@ end
 #  lazy_append(a,b)
 #end
 
-function get_cell_shapefuns(trian::AppendedTriangulation)
```
fverdugo, line 73: Here (in the `Geometry` mdoule) the name `get_cell_shapefuns` was already correct, since the result is the array of cell-wise shape functions.

Revert the renaming in this module and add `import Gridap.Geometry: get_cell_shapefuns` in `FESpaces.jl`

test/CellDataTests/CellDofsTests.jl
```diff
@@ -17,7 +17,7 @@ model = simplexify(CartesianDiscreteModel(domain,cells))
 
 trian = Triangulation(model)
 
-v = GenericCellField(get_cell_shapefuns(trian),trian,ReferenceDomain())
```
fverdugo, line 20: Idem as before, here the old name was correct!

fverdugo (changes requested):
Hi @amartinhuertas !

Some more comments! 

I found that the name `get_cell_shapefuns` was already correct in the `Geometry` module (the new name `get_fe_basis` does not make sense in this context).

See inlined-comments.

src/FESpaces/DivConformingFESpaces.jl
santiagobadia: Yes this is enough because no conformity is to be imposed

src/Geometry/AppendedTriangulations.jl
amartinhuertas: ok, good call. Thanks for detecting this out!

amartinhuertas:
> Some more comments!

Done, ready to be accepted! (once the tests are passing, I will post a comment when that happens)

amartinhuertas:
> Done, ready to be accepted! (once the tests are passing, I will post a comment when that happens)

All tests passing! Thanks @fverdugo and @santiagobadia for your support! I think we can now accept the PR.

fverdugo (approved):

Issue #576, opened 2021-04-11
Generalise Tensor Operations


Omega-xyZac:
This is more of a wish-list item/enhancement: 
It would be great to try and generalise tensor operations so that we don't have to keep introducing new operators. 

An example of such an operation is a operating on tensors that that are defined on our space (I've forgotten the data type that we get here). Eg. $\int_\Omega A_{ijkl}B_{il}~d\Omega. Of course, we could just define a new operation which does that tensor contraction, but it would be nice if we didn't have to do this for every (new) operation.

I believe TensorOperations.jl something like this for nd-arrays. Eg.
```
α=randn()
A=randn(5,5,5,5,5,5)
B=randn(5,5,5)
C=randn(5,5,5)
D=zeros(5,5,5)
@tensor begin
    D[a,b,c] = A[a,e,f,c,f,g]*B[g,b,e] + α*C[c,a,b]
    E[a,b,c] := A[a,e,f,c,f,g]*B[g,b,e] + α*C[c,a,b]
end
```
but obviously this isn't compatible with Gridap.

In Gridap, it would be great if we could do something like
```
∫(A[i,j,k,l]*B[i,l])dΩ 
```
or even
```
C = operate(:contract, A, B, [i,j,k,l], [i,l]) # where i j k l might need to chars or something
∫(C)dΩ
```

Omega-xyZac:
Alternately, if we can efficiently convert to an array type from a general MultiValue type, we could use TensorOperations.jl or another package to do the contraction.

Omega-xyZac:
Ok. We can add some methods to get SArray types for 3-tensors and sym 4-tensors:
```
@generated function _SymFourthOrderTensorValue_to_array(arg::SymFourthOrderTensorValue{D,T}) where {D,T}
  str = ""
  for l in 1:D
    for k in 1:D
      for j in 1:D
        for i in 1:D
          p = _4d_sym_tensor_linear_index(D,i,j,k,l)
          str *= "arg.data[$p], "
        end
      end
    end
  end
  Meta.parse("SArray{Tuple{D,D,D,D},T}(($str))")
end

get_array_new(A::SymFourthOrderTensorValue{D,T}) where {D,T} = _SymFourthOrderTensorValue_to_array(A)
```
and
```
get_array_new(A::ThirdOrderTensorValue{D1,D2,D3,T}) where {D1,D2,D3,T} = SArray{Tuple{D1,D2,D3},T}(Tuple(A))
```
Tested against
[...]

PR #580, opened 2021-04-15
New general tensor operations methods using TensorOperations.jl

Omega-xyZac:
Discussion for this PR can be found in #576. 

Summary:
- We leverage TensorOperations.jl to compute general tensor contractions between Gridap tensor types and Gridap `CellField` tensor types.
- New `get_array` methods are added for `SymFourthOrderTensorValue` and `ThirdOrderTensorValue`.
- A new `contract` method is added to compute the contraction between `MultiValue` types given indices for each type and the result. The output of which is a `MultiValue` type IF the `ndims` of the result from TensorOperations.jl is 1,..,4. 5-tensors etc. will be returned as Arrays.
- A new `GTensor` macro was created co compute general contractions between various Gridap types. This uses a similar notation to TensorOperations.jl. Eg.
```
t1 = ThirdOrderTensorValue(1:27 ...)
t2 = SymFourthOrderTensorValue(1:36 ...)
@GTensor out[i,l,m]=t1[i,j,k]*t2[j,k,l,m]
```
Note: `GTensor` writes to a variable out.
- The implementation of `GTensor` works by creating temporary operations given the provided indices etc.
- Testing was added for both `contract` and `GTensor` in the `OperationsTests.jl` file.

Assumptions:
- `contract` assumes that the output of a 4-tensor is symmetric to follow Gridap typing.


Omega-xyZac:
I'm not entirely sure how we update dependencies?

Omega-xyZac:
It appears the failure is due to failed precompile of TensorOperations.jl. I'm probably doing something wrong in Project.toml.

Omega-xyZac:
This is very frustrating. Code works fine in my modules. Fully tested. Hopefully, you guys can find the issue with the checks. It is something to do with importing `CellData` for `CellField`.

PR #602, opened 2021-05-28
Minor fix in evaluate!(cache,k::TransposeMap,f::ArrayBlock) signature

amartinhuertas:


src/Fields/ArrayBlocks.jl
```diff
@@ -400,7 +400,7 @@ function  return_cache(k::TransposeMap,f::ArrayBlock{A,1} where A)
   ArrayBlock(g,collect(transpose(f.touched))),l
 end
 
-function  evaluate!(cache,k::TransposeMap,f::ArrayBlock)
+function  evaluate!(cache,k::TransposeMap,f::ArrayBlock{A,1} where A)
```
fverdugo, line 403: This makes sense, it is a good thing that the signatures of `return_cache` and `evaluate!` are consistent one with the other.

fverdugo (approved):

PR #605, opened 2021-05-31
Compressed assembly

fverdugo:


fverdugo:
@pmartorell @ericneiva this migth be relevant to you.

PR #623, opened 2021-06-21
Extended functions for ArrayBlock (II)

ericneiva:
Hi @fverdugo,

I have extended `setsize!` for ArrayBlocks and `_compress!` for `Tuple{ArrayBlock,ArrayBlock}`.

Bimaterial tests in `GridapEmbedded#gridap_v0.16` pass, if I apply these changes.

Please, review and accept the merge request, whenever you have some time. Thanks!

### DIFF
--- a/src/Fields/ArrayBlocks.jl
+++ b/src/Fields/ArrayBlocks.jl
@@ -152,6 +152,18 @@ function evaluate!(cache,::typeof(unwrap_cached_array),a::ArrayBlock)
   r
 end
 
+Arrays.setsize!(a::ArrayBlock,s::NTuple{N,<:Integer}) where {N} = @notimplemented
+
+function Arrays.setsize!(a::ArrayBlock,b::ArrayBlock)
+  @assert size(a) == size(b)
+  for i in eachindex(a.array)
+    if a.touched[i]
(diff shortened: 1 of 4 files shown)


src/Fields/ArrayBlocks.jl
```diff
@@ -152,6 +152,18 @@ function evaluate!(cache,::typeof(unwrap_cached_array),a::ArrayBlock)
   r
 end
 
+Arrays.setsize!(a::ArrayBlock,s::NTuple{N,<:Integer}) where {N} = @notimplemented
+
+function Arrays.setsize!(a::ArrayBlock,b::ArrayBlock)
```
fverdugo, line 157: I am afraid that this would not work when we have nested blocks. I have to think how to implement this.

fverdugo (changes requested):

src/Fields/ArrayBlocks.jl
ericneiva: I see, I had not thought about nested blocks...

fverdugo: It seems that are tests are about to pass with the new implementation

PR #628, opened 2021-07-07
Evaluate any monomial basis at single point

Balaje:
Hi @fverdugo 

Here is the PR to add the feature to evaluate any monomial basis at a single point. @ericneiva mentioned you are on vacation, so you can have a look at it whenever you get back! 🙂


fverdugo (approved):
Elegant fix!

PR #636, opened 2021-08-12
div(v)*p and q*div(u) trick in reference space high level API for high order geometric mappings

amartinhuertas:
Hi @fverdugo,

I have come up with a proposal to support the following trick for the implementation of this term (Darcy+RT FEs)

Screenshot from 2021-08-12 15-11-47

This will let us by-pass the current limitation that we have in the computation of the divergence operator. (i..e, Gridap assumes that the mapping is affine, and thus that the Jacobian is constant within each cell). This prevents the usage of, e.g., quadrilateral bilinear disctrizations of the sphere using the so-called cubed-sphere mesh, but even worse, it prevents curved elements at all. 

In code, you may se the new term in action here: https://github.com/gridap/Gridap.jl/commit/b989735901516007b1775486348fc8b6a53f0591#diff-37219e6468e24d7c4b29903834b894aa426a861aa49ef64122068cbd47ceb5daR24

Concerns to take into account when reviewing:

* Is there any other possible approach to implementation? If yes, pros/cons?
* I have tried to do my best to place the difference pieces of code within the source code hierarchy. Review this, please. 
* The `DIV()` operator is currently only defined for instances of type `SingleFieldFEBasis`. Should it be defined for any other objects?
* Is composability ok with the way `ReferenceTriangulation` triangulation is implemented?
*  `ReferenceTriangulation` `TriangulationStyle` [here](https://github.com/gridap/Gridap.jl/commit/b989735901516007b1775486348fc8b6a53f0591#diff-2815e471f041be05270b00dfa8c373fdce9c1407342bec66ec513f7c4a906408R21) correct?

TO-DO:
* [x] Modify `NEWS.md` once we agree on the approach and the details. 

PS: @santiagobadia already took a look at the approach to implementation, he thinks it is reasonable, and agreed in the details.


fverdugo:
Which is the math definition of the new operator DIV? It is more than the div wrt reference coordinates right?

amartinhuertas:
There is one paper by botchev ... Rehabilitation of lowest order ... That defines a discrete variant of the physical div as the normal flux integrated over the boundary divided by the element measure. I am not sure of the connections of that DIV operator and the one at hand (divergence in reference space)

fverdugo:
I want to understand why this new feature does not fit well in our current design. 

Perhaps it is related with the fact that the high level API always refers to the physical space (i.e. the gradient and div are always wrt the physical coordinate), but in some situations you also want to write integrals of gradients/divs in the reference cell with the high level API (which is not supported now)

If this is the reason, perhaps we need to introduce a  `ReferenceMeasure` instead of a `ReferenceTriangulation` Or instead of a `ReferenceMeasure` parametrize `Measure` with `DomainStyle`. E.g.,:
```julia
dΩ = Measure(Ω) # PhysicalDomain() by default
dω = Measure(Ω,ReferenceDomain())
∫( ∇(u)⋅∇(v) )dΩ # Gradients wrt physical coordinate
∫( ∇(u)⋅∇(v) )dω # Gradients wrt reference coordinate
```
I like this API, but not sure how difficult it is to implement it.

fverdugo:
> There is one paper by botchev ... Rehabilitation of lowest order ... That defines a discrete variant of the physical div as the normal flux integrated over the boundary divided by the element measure. I am not sure of the connections of that DIV operator and the one at hand (divergence in reference space)

Ok! So the `DIV` operator is not a trick. it is something with a math definition, right? So it is perfectly fine to add it.

Which is the "tricky" part of the PR then? The `ReferenceTriangulation` ?

amartinhuertas:
Yes the reason you say. The high level API is designed to always perform the change of domain from physical to reference space. It computes the Jacobian of the mapping from reference to physical and uses it to transform the differential integral measure

fverdugo:
OK! then the part we have to work a bit more is the `ReferenceTriangulation`

amartinhuertas:
I called it trick because by exploiting the particular definition of the pull back of RT shape function it lets you compute that integral without having to compute the divergence in physical space which is complicated for non affine geometries due to the piola map

amartinhuertas:
> I like this API, but not sure how difficult it is to implement it.

Ok, I already remove `ReferenceTriangulation` and implemented the API you proposed. See:

https://github.com/gridap/Gridap.jl/pull/636/commits/a6cc908da289ce453413b069d2bc4151fcc628b6#diff-37219e6468e24d7c4b29903834b894aa426a861aa49ef64122068cbd47ceb5daR20

BTW, another different (benefit?) in the proposed solution is that we don't have different entries in the dictionary within `DomainConstribution` for each kind of domain (i.e., reference versus physical). I guess that mixing in the same lazy_map both contributions one might open the door for further optimizations.

src/CellData/CellQuadratures.jl
```diff
@@ -3,56 +3,93 @@
...
-struct CellQuadrature{DS} <: CellDatum
+struct CellQuadrature{DDS,IDS} <: CellDatum
   cell_quad::AbstractArray{<:Quadrature}
   cell_point::AbstractArray{<:AbstractArray{<:Point}}
   cell_weight::AbstractArray{<:AbstractArray{<:Real}}
   trian::Triangulation
-  domain_style::DS
+  data_domain_style::DDS
+  integration_domain_style::IDS
 end
```
fverdugo, line 13: The default constructor of the struct has changed. Can you add an extra constructor with the old signature? and perhaps calling it in the tests? Just to be backwards compatible, I.e.,

```julia
function CellQuadrature(
  cell_quad::AbstractArray{<:Quadrature},
  cell_point::AbstractArray{<:AbstractArray{<:Point}},
  cell_weight::AbstractArray{<:AbstractArray{<:Real}},
  trian::Triangulation,
  data_domain_style::DomainStyle)
  CellQuadrature(cell_quad,cell_point,cell_weigth,trian,data_domain_style,PhysicalDomain())
end
```

fverdugo (changes requested):
Hi @amartinhuertas thanks for the changes.

Just a very minor comment and its ready to merge!

src/CellData/CellQuadratures.jl
amartinhuertas: Yes! I will do that soon (I am in the middle of another task). 

I guess that the current set of tests do not fail because they are not calling to the (old) default constructtor, right?

fverdugo:
> I guess that the current set of tests do not fail because they are not calling to the (old) default constructtor, right?

Yes....

src/CellData/CellQuadratures.jl
amartinhuertas: > Yes! I will do that soon (I am in the middle of another task).

Done!

fverdugo (approved):

amartinhuertas:
@fverdugo ...  the PR can be accepted, right?

amartinhuertas:
Thanks!

Issue #637, opened 2021-08-12
Check efficiency of RT implementation


fverdugo:
By looking at https://github.com/gridap/Gridap.jl/blob/67e5ee146ee0a3731db30c79ae2c5144c160c1b1/src/FESpaces/DivConformingFESpaces.jl#L116

I would say that we are reevaluating the reference polynomials each time we visit a new cell.

fverdugo:
I think that we need to replace
``` julia
    lazy_map(_transform_rt_shapefuns,
             cell_reffe,
             get_cell_map(Triangulation(model)),
             lazy_map(Broadcasting(constant_field), sign_flip))
``` 

by

``` julia
cell_fs = lazy_map(get_shape_functions,cell_reffe)
lazy_map(ContraPiolaMap(),cell_fs,
             get_cell_map(Triangulation(model)),
             lazy_map(Broadcasting(constant_field), sign_flip))
```
and remove `_transform_rt_shapefuns`.

Perhaps, Idem with the celldofs

amartinhuertas:
## Base line commit 67e5ee146ee0a3731db30c79ae2c5144c160c1b1  (2D Darcy on a box) 

Profile data attached below

```
10×10 DataFrame
 Row │ Assembly RT [s]  n      k      d      Total RT [s]  num_cells  l2_err_norm  Preassembly RT [s]  experiment      Solve RT [s] 
     │ Float64          Int64  Int64  Int64  Float64       Int64      Float64      Float64             String          Float64      
─────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │        0.33423      10      0      2      0.337986        100  9.02765e-15          0.00177146  solve_darcy_rt    0.00198408
   2 │        0.366117     20      0      2      0.378329        400  3.08589e-14          0.00284878  solve_darcy_rt    0.00816082
   3 │        0.374344     30      0      2      0.403508        900  3.03932e-14          0.00273725  solve_darcy_rt    0.0226722
   4 │        0.387381     40      0      2      0.430464       1600  7.15834e-14          0.00387972  solve_darcy_rt    0.0282138
   5 │        0.4075       50      0      2      0.47579        2500  9.25386e-14          0.00481167  solve_darcy_rt    0.0538609
   6 │        0.432286     60      0      2      0.534182       3600  1.78746e-13          0.00550788  solve_darcy_rt    0.0934964
[...]

PR #638, opened 2021-08-13
Optimizing conforming rt fes

amartinhuertas:
see issue https://github.com/gridap/Gridap.jl/issues/637


amartinhuertas:
Hi @fverdugo ... the tests in this PR failed due to VERY strange error:

https://github.com/gridap/Gridap.jl/pull/638/checks?check_run_id=3321610616#step:6:131

In particular, the symbol `_transform_rt_shapefuns` no longer exists in the source code! (I grep it, and there is no such symbol in the whole source directory). 

Can it bit some sort of dirty cache in Github actions or something like that? Have you ever find something similar?

amartinhuertas:
OK ... this PR is already for review, up to solving the following:

1. https://github.com/gridap/Gridap.jl/pull/638#issuecomment-898423224
2. https://github.com/gridap/Gridap.jl/issues/637#issuecomment-898387437
3. https://github.com/gridap/Gridap.jl/issues/637#issuecomment-898439766

fverdugo:
checks are not displayed. Perhaps a dummy commit to re-run them?

amartinhuertas:
> checks are not displayed. Perhaps a dummy commit to re-run them?

Ok. I have pushed a new commit and the issue seems to have disappear ... weird !

amartinhuertas:
@fverdugo ... FYI ... this PR is ready for review

src/Fields/ArrayBlocks.jl
```diff
@@ -296,6 +296,14 @@ function lazy_map(::typeof(evaluate),a::LazyArray{<:Fill{<:BlockMap}},x::Abstrac
   lazy_map(k,args...)
 end
 
+function lazy_map(k::Broadcasting,a::LazyArray{<:Fill{<:BlockMap}})
```
fverdugo, line 299: This is the only think that I am not sure about in this PR. 

Perhaps replacing `Broadcasting` by `BroadcastingFieldOpMap` makes more sense.

fverdugo (changes requested):

src/Fields/ArrayBlocks.jl
amartinhuertas: > This is the only part that I am not sure about in this PR.

Ok, this is what I was also wondering; see https://github.com/gridap/Gridap.jl/issues/637#issuecomment-898387437

> Perhaps replacing Broadcasting by BroadcastingFieldOpMap makes more sense.

To put you in context (if you were not already), this `lazy_map` overload is currently triggered from this line

https://github.com/gridap/Gridap.jl/blob/master/src/ReferenceFEs/RaviartThomasRefFEs.jl#L392

If I understand correctly, you want to defer this "change of order" of operations till evaluation, right? Apart from consistency with other parts of the code, what are the benefit of this? (for my understanding).

amartinhuertas:
Also @fverdugo ... do u have any insight on this question? https://github.com/gridap/Gridap.jl/issues/637#issuecomment-898816636

fverdugo:
My concern is that this
```julia
function lazy_map(k::Broadcasting,a::LazyArray{<:Fill{<:BlockMap}})
  args = map(i->lazy_map(k,i),a.args)
  bm = a.maps.value
  lazy_map(bm,args...)
end
```
is only valid when `k` represents a  linear operation.

Unfortunately. `BroadcastingFieldOp` will not fix it since you are broadcasting the gradient.

I would say that `v` in https://github.com/gridap/Gridap.jl/blob/7ec790dd3ed2d5b1ca28327843e58df52a7e44fa/src/ReferenceFEs/RaviartThomasRefFEs.jl#L392
is not "multifield" yet. Thus, `ArrayBlock` and `BlockMap` should not be involved here.  

If `v` is "multifield" for some reason. I would try to fix this and then you will not need to introduce the new method dispatch def above.

fverdugo:
> Also @fverdugo ... do u have any insight on this question? #637 (comment)

Assembly routines can allocate large amounts of memory (e.g. intermediate sparse representations of your matrix). The explicit call to `gc()` is to free them as soon as possible. 

In any case, after commenting the `gc` call, there is still a large plateau in https://github.com/gridap/Gridap.jl/issues/637#issuecomment-898830041

Do you know what is causing this? perhaps is also related with `gc`

amartinhuertas:
> Assembly routines can allocate large amounts of memory (e.g. intermediate sparse representations of your matrix). The explicit call to gc() is to free them as soon as possible.

Ok, this is what I thought. Did you observe the positive impact of gc in memory consumption? Have you studied this in detail? Perhaps (this is speculation, I am not an expert in Julia internals) by forcing the gc to enter in a particular moment in time, instead of letting it decide when it is best to enter, is causing performance issues. 

>In any case, after commenting the gc call, there is still a large plateau in #637 (comment)
> Do you know what is causing this? perhaps is also related with gc

Ok, that is a good observation. This plateu was also present in the flame graphs previous to deactivating gc and occurs during the solve stage. As you say, looking at the RT solve column before and after gc, there has been an increase of the computational time of RT solve, but this is increase is not as significant as the one in Assembly when we have explicit gc.

amartinhuertas:
Looking at the Total column helps in determining the best compromise.

amartinhuertas:
> is only valid when k represents a linear operation.

by "linear" you mean k(x+y)=k(x)+k(y) and k(alpha*x)=alpha*k(x) ? 

I do not see the relantionship among being linear in this sense and the impossibility/uncorrectness of defining the result of broadcasting an operation to a vector block as the vector block of broadcasting the operation to its blocks. (Sure I am missing something.) 

> If v is "multifield" for some reason. I would try to fix this and then you will not need to introduce the new method dispatch def above.

Ok, let me explore this and I will come back to you if I find a solution.

fverdugo:
If k is not linear linear, i.e., k(a*x) != a*k(x), zero blocks in the input can result in non-zero blocks in the output.

amartinhuertas:
>I would say that v in XXX is not "multifield" yet. Thus, ArrayBlock and BlockMap should not be involved here.
If v is "multifield" for some reason. I would try to fix this and then you will not need to introduce the new method dispatch def above.

@fverdugo ... I have been thinking carefully about this comment. While we may try to achieve this, let me stress that the comment is **NOT** consistent with what Gridap is currently doing in analagous situations. To illustrate my point, I wrote the following MWE:

```julia
using Gridap
using Gridap.CellData
n = 3
domain = (0,1,0,1)
partition = (n,n)
model = simplexify(CartesianDiscreteModel(domain, partition))

order = 1
reffeᵤ = ReferenceFE(lagrangian,Float64,order)
reffeₚ = ReferenceFE(lagrangian,Float64,order)
V = TestFESpace(model,reffeᵤ,conformity=:H1)
Q = TestFESpace(model,reffeₚ,conformity=:H1)
Y = MultiFieldFESpace([V,Q])

v=get_fe_basis(V)
y=get_fe_basis(Y)
vb,_=y
∇(vb)
```

and debugged carefully the last line. This line triggers the following line:

https://github.com/gridap/Gridap.jl/blob/f1ad18665a53e71fc5b5e284ca9311dd6f48b983/src/CellData/CellFields.jl#L428

which in turns triggers:

https://github.com/gridap/Gridap.jl/blob/f1ad18665a53e71fc5b5e284ca9311dd6f48b983/src/Fields/ArrayBlocks.jl#L459

Perhaps I am missing something, but I do not see how come this is different from what I am trying to achieve with the method dispatch I have introduced. The difference among 

https://github.com/gridap/Gridap.jl/blob/f1ad18665a53e71fc5b5e284ca9311dd6f48b983/src/CellData/CellFields.jl#L428

and my particular case, is that in the former the array to which we apply the `Broadcasting(gradient)` operation is a Fill array of VectorBlock instances, while in the latter, the array to which we apply `Broadcasting(gradient)` operation is a LazyArray with `BlockMap` as a kernel. I guess that this is a new case that had not been considered before, and this is why we need the definition for optimization purposes. 

> If k is not linear linear, i.e., k(ax) != ak(x), zero blocks in the input can result in non-zero blocks in the output.

See my comment above. Isn't 

https://github.com/gridap/Gridap.jl/blob/f1ad18665a53e71fc5b5e284ca9311dd6f48b983/src/Fields/ArrayBlocks.jl#L459

violating the property you are referring to?

amartinhuertas:
> To put you in context (if you were not already), this lazy_map overload is currently triggered from this line

> https://github.com/gridap/Gridap.jl/blob/master/src/ReferenceFEs/RaviartThomasRefFEs.jl#L392

> If I understand correctly, you want to defer this "change of order" of operations till evaluation, right? Apart from consistency  with other parts of the code, what are the benefit of this? (for my understanding).

Ok. I have now realized (I have run the debugger to double check) that this statement I am quoting is WRONG (sorry about that!).

In particular, the new `lazy_map` overload is triggered from this line:

https://github.com/gridap/Gridap.jl/blob/f1ad18665a53e71fc5b5e284ca9311dd6f48b983/src/CellData/CellFields.jl#L428

Here `get_data(a)` is a LazyArray + `BlockMap` kernel. (instead of a Fill array of VectorBlocks as with Lagrangian FEs).

fverdugo:
> In particular, the new lazy_map overload is triggered from this line:

Ok I see! Then we can add the new method overload you proposed but restricted to the gradient to be on the safe side (i.e. to avoid using it with non linear kernels `k`).  Do you think it's enough for your use case?

```julia
function lazy_map(k::Broadcasting{typeof(gradient)},a::LazyArray{<:Fill{<:BlockMap}})
  args = map(i->lazy_map(k,i),a.args)
  bm = a.maps.value
  lazy_map(bm,args...)
end
```

amartinhuertas:
> Then we can add the new method overload you proposed but restricted to the gradient to be on the safe side (i.e. to avoid using it with non linear kernels k)

Ok, deal. 

Is there any example of a non-linear kernel in Gridap.jl used in conjunction with `Broadcasting` or it is a future possibility? 

I want to make sure I have understood the point. (to avoid introducing the same issue in the future). 

> Do you think it's enough for your use case?

Yes. (I have double checked). 

It remains to decide what to do with the explicit call to GC(). It is currently commented in this PR.

fverdugo:
We can remove the call to `gc()` as you propose.  This can be readily merged, right?

amartinhuertas:
I think so. It can be merged.

fverdugo (approved):

PR #639, opened 2021-08-13
Easy improvements in interpolatable

fverdugo:


fverdugo:
@Balaje  I have done some easy performance improvements (see the diff).

Now the situation is as follows:

Given this simple example (run it with --check-bounds=no):
```julia
using Gridap
using Gridap.Arrays
using Gridap.Fields
using Gridap.FESpaces
using Gridap.CellData
using BenchmarkTools
using ProfileView

k = 1
f(x) = x[1] + x[2]
reffe = ReferenceFE(lagrangian,Float64,k)

# Mesh 1
n_1 = 100
model_1 = CartesianDiscreteModel((0,1,0,1),(n_1,n_1))
V_1 = FESpace(model_1,reffe)

# Mesh 2
n_2 = 200
model_2 = CartesianDiscreteModel((0,1,0,1),(n_2,n_2))
V_2 = FESpace(model_2,reffe)

# from 1 to 2
vh_1 = interpolate(f,V_1)
k = Interpolable(vh_1)
vh_2 = interpolate(k,V_2)
@profview interpolate(k,V_2)
```

the flame graph looks like this:
Screenshot from 2021-08-13 15-55-37

which suggest that the problem is the call to `nn`https://github.com/gridap/Gridap.jl/blob/ac24cf4303dcda9d9f8f24fb5c19726790b621b0/src/CellData/CellFields.jl#L273

I don't know if we can fix this from the Gridap side. Perhaps one needs to fix `nn`.

PR #648, opened 2021-08-19
Fixing naming convention for BoundaryDiscreteModel.jl

amartinhuertas:


amartinhuertas:
This PR can be accepted straight away once the tests pass. It is just a raming change.

Issue #646, opened 2021-08-19
Current approach to manifold mesh generation


amartinhuertas:
Hi @fverdugo, @santiagobadia, 

I have spent some time looking at the approach we are following to generate discretizations of a manifold, to represent them in `Gridap.Geometry`, and to compute the manifold mesh outward unit normal (I think @fverdugo has mentioned that the current approach is temporary, not sure if the following is among the causes.) 

While this approach is clearly functional (e.g., it has let us solve PDE problems on the cubed sphere), I am not sure about the following:

 * does it faithfully represent the underlying mathematical abstraction?
 * is it general enough?
 * it it efficient enough?
 
if the answer to any of these questions is "not enough", this may require some more profound thinking, that perhaps we can incorporate in the future to the very much needed `Gridap.Geometry` refactorization that we are anticipating (for different reasons, though).

(The following is inferred from my understanding of what I have read in the code, if I am not right, confused, etc., I am sorry ... let me please know if this is the case)

The current approach relies on `BoundaryDiscreteModel`. This data type is created from a volume discrete model. Taking profit of this, `Triangulation(::BoundaryDiscreteModel)`  returns a `BoundaryTriangulation` instance. Using this latter instance you can compute the manifold's outward unit normal using the function `get_normal_vector(::BoundaryTriangulation)`, **which uses the inverse of the volume mesh cell Jacobian, the normal in reference space, etc.** So far so good. 

However, I see the following issues:

1. With `BoundaryDiscreteModel/BoundaryTriangulation`, by conception,  there is a volume mesh in the background. However, my understanding is that  manifolds can exist without a volume in the background, e.g., you might read a 2D manifold mesh embedded in 3D directly from data files, e.g., in STL or GeoMesh format.   
2. Even if in the particular case of `BoundaryDiscreteModel`, there is a volume mesh in the background,  we may compute the normal in a cheaper way from the computation POV (e.g., computation of inverse of volume mesh map no longer necessary). 

[...]

PR #649, opened 2021-08-25
Fixing normal skeleton triangulation for manifolds

amartinhuertas:
Related to this comment https://github.com/gridap/Gridap.jl/issues/646#issuecomment-901727279


amartinhuertas:
Question: might we be applying the same (wrong) identity at other parts of the code?

fverdugo:
Hi @amartinhuertas thanks for noting this particular case!

fverdugo (approved):

PR #651, opened 2021-08-31
Fixing bug in boundary discrete model

amartinhuertas:


fverdugo:
Hi @amartinhuertas, can you briefly comment on what the problem was?

amartinhuertas:
> Hi @amartinhuertas, can you briefly comment on what the problem was?

Sure. The previous-to-the-fix function aimed to preserve the relative order in the numbering of the objects of the background model in the numbering of the objects filtered for the BoundaryDiscreteModel. This is not necessarily OK for the objects of the highest dimension, i.e., the cells in the `BoundaryDiscreteModel`, as you may want to change the numbering of the cells in `BoundaryDiscreteModel` (i.e., their relative order in the background model).

amartinhuertas:
@fverdugo ... FYI ... the following test (that I added on porpuse)

https://github.com/gridap/Gridap.jl/pull/651/files#diff-acafa9fbe3103748c87293ace08fb9618d888c62584ecac02e694f535a888badR63

was broken before the bug fix

fverdugo:
> Sure. The previous-to-the-fix function aimed to preserve the relative order in the numbering of the objects of the background model in the numbering of the objects filtered for the BoundaryDiscreteModel. This is not necessarily OK for the objects of the highest dimension, i.e., the cells in the BoundaryDiscreteModel, as you may want to change the numbering of the cells in BoundaryDiscreteModel (i.e., their relative order in the background model).

Ok. In any case BoundaryDiscreteModel is likely to be repaced by something else in the future to simplify the handling of mixed dimensional PDEs.

amartinhuertas:
> In any case BoundaryDiscreteModel is likely to be repaced by something else in the future to simplify the handling of mixed dimensional PDEs.

Yes, please, :-)

Issue #660, opened 2021-09-24
Pending tasks after Geometry module refactoring.


fverdugo:
- [ ] Clean up: remove old files, do not export deleted functions etc.
- [ ] Implement a compression mechanism for unfitted FEM in the new version
- [ ] Caching the result of `get_active_model` for instances of `BodyFittedTriangulation`, `BoundaryTriangulation`
- [ ] A more elegant way of computing the restriction of the restriction of a triangulation in the user driver

fverdugo:
WIP in branch https://github.com/gridap/Gridap.jl/tree/geometry

PR #662, opened 2021-09-27
Geometry

fverdugo:
Major refactoring on how data is transferred between different triangulations.

- Allows to implement mixed dimensional PDEs in a more elegant way
- More extensible design that will open the door to include HDG methods.

## Changed

The `Triangulation` interface has changed. Now a `Triangulation` object is understood as computational domain defined from a `DiscreteModel`. I.e., from the same FE mesh (the discrete model), we can extract different domains (several triangulations) like the boundary, the interior, or some portion of the interior or the boundary. These can be potentially used both for intergration (like before) but also for interpolation (except for unfitted triangulations). In consequence, types like `BoundaryDiscreteModel` or `ExtendedDiscreteModel` are no longer needed.


fverdugo:
Some remaining tasks associated with this PR are detailed in #660

Issue #584, opened 2021-04-29
Automatic differentiation in multifield with fields of different num dofs.


fverdugo:
This has been broken after PR #583 and should be fixed.

fverdugo:
Hi @amartinhuertas,

In order to fix this, we need the methods you implemented for merging an ArrayBlock into a standard Julia Array. 

Can you isolate this implementation and open a PR?

Once we have this, fixing the autodiff for multi-field pdes should be easy.

amartinhuertas:
> Can you isolate this implementation and open a PR?

Ok. I will try to do it. I cannot guarantee to have it very soon, though, I have many other tasks on the queue.

amartinhuertas:
> Once we have this, fixing the autodiff for multi-field pdes should be easy.

Hi, @fverdugo ... do you know what else do we need to solve this issue? I can continue contributing ...

amartinhuertas:
> The idea is to merge the blocks of the residual and the dof values before calling AD. Now, we are trying to call AD for each block separately that seems to fail when you merge the blocks back due to a mismatch in the chunk size of the underlying dual numbers.

Ok. I will try to take a look (whenever I have some time) at the particular part in which this is carried out ...

PR #687, opened 2021-10-22
Solving ad multifield

amartinhuertas:
Related to issue https://github.com/gridap/Gridap.jl/issues/584

Hi @fverdugo, 

find attached a first approach to solve the issue related to the `gradient` of a FE functional evaluated at a multifield FE function with different number of DoFs per field. Once we are happy with the current solution for `gradient`, I will extend this solution to `jacobian`. Agreed?


tmigot:
Hi @amartinhuertas and @fverdugo ! Thanks for fixing this issue :). I was wondering if you also planned to add the MultiField `hessian` function?

PR #701, opened 2021-10-28
Operation tree optimization

oriolcg:
Hi @fverdugo , 

This is the implementation of the tree optimization. I haven't spent much time assessing the performance improvements, but comparing at the Drivers tests, it looks like that the optimized version is a bit slower. Do you see any point of improvement in the implementation?


oriolcg:
Together with @ericneiva we made some "optimizations", but still seeing more time in the Drivers tests. We need to further investigate this issue before merging. We believe that there are some extra allocations when looping over the dictionary entries to check if the branch is the same.

fverdugo:
Thanks for this work!

We also need to see if this indeed reduces the run times significantly in some complex example.

Project.toml
```diff
@@ -21,6 +21,7 @@ NLsolve = "2774e3e8-f4cf-5e23-947b-6d7e65073b56"
 NearestNeighbors = "b8a86587-4115-5ab1-83bc-aa920d37bbce"
 QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
fverdugo, line 24: Please do not add `Revise`. Having it installed in your global environment should be enough.

```
$ julia
(@1.6) pkg> add Revise 
```

Then 
```
$ julia --project=whatever/
julia> using Revise
```
should work.

fverdugo (changes requested):
@oriolcg thanks for the update! See the minor comment.

We keep this open until we have some confirmation that the optimization has a positive effect.

Project.toml
ericneiva: Done

ericneiva:
Hi, @fverdugo, we are positive that subbranches that are equal but receive different object ids are indeed detected via `same_branch`. However, the optimisation has not any positive effect, on the contrary, more allocations. We believe, according to benchmark tools output, that there are some extra allocations when looping over the dictionary to check if the subbranch is already there https://github.com/gridap/Gridap.jl/blob/operation_tree_optimization/src/Arrays/LazyArrays.jl#L154, but we do not still know exactly where, maybe when extracting the caches that are in the dictionary.

fverdugo:
These extra allocations perhaps do not grow with the number of cells hopefully. Have you profiled some large example with —check-bounds=no -O3? I can join the next coding session and build a bench together

ericneiva:
> Have you profiled some large example with —check-bounds=no -O3?
We plan to do it next Wednesday afternoon.
> I can join the next coding session and build a bench together
That would be great!

fverdugo (approved):

PR #704, opened 2021-10-29
Linear reference grid for simplex reference FEs and polytopes

pmartorell:
Dear @fverdugo,

this PR implements the construction of a linear reference grid for simplex reference FEs and polytopes.

This functionality is accessed through `compute_reference_grid(reffe::ReferenceFE,nelem::Integer)` and `compute_reference_grid(poly::Polytope,nelem::Integer)`. Which return an `UnstructuredGrid` of simplices. or a `CartesianGrid` in the case of n-cubes. This may lead to type instability.


fverdugo (approved):
Hi @pmartorell,

Thanks for this PR! This was an important missing part in the code. Nice that finally we have it!

PR #715, opened 2021-11-21
Fixes test_broken in GridapDistributed

amartinhuertas:
Fixes PR https://github.com/gridap/GridapDistributed.jl/pull/61


fverdugo:
Reproducer:

```julia
module TMP

using Gridap
inds = Int32[1,4,3]
a = rand(10)
sa = Gridap.Arrays.SubVector(a,1,5)
vsa = view(sa,inds)

b = rand(10)
vb = view(b,inds)

x = Base.broadcasted(+,vb,vsa)
Base.materialize!(vb,x)

end # module

```

PR #718, opened 2021-11-26
Using nospecialize during assembly

amartinhuertas:
to get a grasp of the consequences of this in Github actions ...

### DIFF
--- a/src/FESpaces/SparseMatrixAssemblers.jl
+++ b/src/FESpaces/SparseMatrixAssemblers.jl
@@ -220,6 +220,7 @@ function symbolic_loop_matrix!(A,a::GenericSparseMatrixAssembler,matdata)
 end
 
 @noinline function _symbolic_loop_matrix!(A,caches,cell_rows,cell_cols,mat1)
+  @nospecialize cell_rows, cell_cols
   touch_cache, rows_cache, cols_cache = caches
   touch! = TouchEntriesMap()
   for cell in 1:length(cell_cols)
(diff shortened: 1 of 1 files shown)


fverdugo:
Stokes DG has improved compile times (198 vs 337):
https://github.com/gridap/Gridap.jl/runs/4335204080?check_suite_focus=true#step:7:22
https://github.com/gridap/Gridap.jl/runs/4284779963?check_suite_focus=true#step:7:22

The other tests have not varied so much. Perhaps this is showing that to benefit from the change you need a sufficiently complex weak form.

fverdugo:
@amartinhuertas can you test your killer code with this branch?. If this solves the problem, perhaps we have a sufficiently strong motivation to merge this.

oriolcg:
Hi @fverdugo @amartinhuertas, I tried this branch for the Embedded ALE FSI test that I have. This code used to take 1500-2000 secs to compile. With this branch it takes ~800. Execution times after the first run are of the same order (not properly measured, though)

amartinhuertas:
> @amartinhuertas can you test your killer code with this branch?. If this solves the problem, perhaps we have a sufficiently strong motivation to merge this.

Definitely a game changer. The first call latency of the killer code has been reduced from more than a dozen of hours to 18 minutes with Julia 1.6.4. Moreover, the same code with Julia 1.5.4 takes round 48 minutes. This latter result does not surprise me, I have observed that with Julia 1.5.4 GridapDistributed/PartitionedArrays takes much longer to compile than with Julia 1.6.4. 

I think we should merge this, once we clarify which is the "optimal" setting for the usage of the nospecialize method. It would be also great to understand the cause of the bottleneck so that we can identify where could arise as well in order parts of the code.

amartinhuertas:
I guess we can abandon this PR, right? Closing.

PR #720, opened 2021-11-30
Workarounds for faster compilation

fverdugo:


fverdugo:
I am not convinced that we want to merge this after my tests reported in issue #723

amartinhuertas:
> I am not convinced that we want to merge this after my tests reported in issue #723

Ok. Closing PR.

Issue #722, opened 2021-11-30
Issues with `TriangulationView`?


amartinhuertas:
Hi @fverdugo, 

I could narrow down a code that reproduces the issue we have started discussing in Slack. 

The code, as suggested, only depends on Gridap, and looks like

```julia
using Gridap
model = simplexify(CartesianDiscreteModel((0,1,0,1),(2,2)))
reffe_lgn = ReferenceFE(lagrangian, Float64, 1)
S = FESpace(model, reffe_lgn; conformity=:H1)
s=get_fe_basis(S)
Ω=Triangulation(model)
grads=gradient(s)
Ωo=Gridap.Geometry.TriangulationView(Ω,[1,2,3])
no=Gridap.CellData.GenericCellField(get_cell_map(Ωo),Ωo,Gridap.CellData.ReferenceDomain())
b = Gridap.CellData._to_common_domain(no,grads)
x = Gridap.CellData._get_cell_points(b...)
y2=Gridap.CellData.get_data(b[1]).args[1].args[1].args[2]
y2[4] # Out of bounds error! Is this expected?
```
Note that i have used `Ωo=Gridap.Geometry.TriangulationView(Ω,[1,2,3])`, instead of `view(Ω,[1,2,3])`. The former is exactly what we have in `GridapGeosciences` when I use the latter with the triangulation resulting from `CubedSphereDiscreteModel`. For completeness, let me copy here the code of `GridapGeosciences` that also exposes the same out of bounds error:

 ```julia
using Gridap
using GridapGeosciences


model = CubedSphereDiscreteModel(1; radius=rₑ)
reffe_lgn = ReferenceFE(lagrangian, Float64, 1)
S = FESpace(model, reffe_lgn; conformity=:H1)
s=get_fe_basis(S)
Ω=Triangulation(model)
grads=gradient(s)
Ωo=view(Ω,[1,2,3])
no=get_normal_vector(Ωo)
b = Gridap.CellData._to_common_domain(no,grads)
x = Gridap.CellData._get_cell_points(b...)
y=Gridap.CellData.get_data(b[1]).args[1].args[1].args[1].args[1].args[2]
y[4]
```

Let me recall that the the out of bounds error is triggered  from the `_have_same_ptrs(g)` Gridap.jl private function. Any clue about this? (1) Is this a BUG? (2) If not, what is wrong? (3) If yes, how do we solve it? 

Thanks!

fverdugo:
Even a more minimal example:

```julia
[...]

PR #724, opened 2021-11-30
Fix issue #722

fverdugo:


fverdugo:
Hi @amartinhuertas. 

Feel free to merge once the tests pass. If you are in a hurry, feel also free to release a new patch version after including some comment in NEWS.md that we have fixed a bug related with `TriangulationView`

Issue #727, opened 2021-12-01
Issue finding cells in interpolate


oriolcg:
Hi @fverdugo, 

there is an error in the algorithm for finding the cell where a given point lies for unstructured grids. This error is manifested when a point lies in an element whose vertices are not the closes vertex to the point. See for instance the following example:

![tmp](https://user-images.githubusercontent.com/9356155/144309790-3026569a-b5e4-4d5d-a8fe-181a8595f7d5.png)
:

One possibility for a workaround would be to find through the neighboring elements if any cell is found in the first pass. But I'm not sure if this is efficient and I don't know how to implement it in the most efficient manner. Any ideas?

fverdugo:
@Balaje any thoughts on this?

Balaje:
Hi @fverdugo @oriolcg ,

I think the best way is to look at the other neighbouring points (I think `knn` does that instead of `nn` https://github.com/KristofferC/NearestNeighbors.jl#k-nearest-neighbor-knn-searches) and see if the correct element is found. I can see that this is what @oriolcg has implemented.

Balaje:
Also, @oriolcg loops over 3 nearest neightbours right now. Can we also have an option for the level of searching, something like what is done here using `searchMethod` https://doc.freefem.org/documentation/finite-element.html#a-fast-finite-element-interpolator?

oriolcg:
Actually, I'm only looping for the first and if not found it goes to second and third. So, keeping aside the time of getting the closest 3 vertices instead of only 1, extra work is only done for the pathologic cases.

PR #729, opened 2021-12-03
Interpolation point error

oriolcg:
Fixing issue #727, closing #727 .

### DIFF
--- a/src/CellData/CellData.jl
+++ b/src/CellData/CellData.jl
@@ -61,6 +61,7 @@ export CellDof
 export get_normal_vector
 export get_cell_measure
 export Interpolable
+export KDTreeSearch
 
 export make_inverse_table
 export compute_cell_points_from_vector_of_points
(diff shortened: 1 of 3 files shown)


fverdugo:
@Balaje are you happy with this?

oriolcg:
@fverdugo, I did an update trying to go in the direction that @Balaje suggested.  I now use the `searchmethod` in `Interpolable` to define an optional argument with the number of nearest vertices (default = 1). This is much cleaner, this optional argument can be changed directly from the user driver (if needed) as follows:

```julia
sm=KDTreeSearch(num_nearest_vertices=10)
uₕ = interpolate_everywhere(Interpolable(uh_0;searchmethod=sm),U)
```

Balaje:
@oriolcg @fverdugo Yes, this is what I had in mind. It gives the user an option to provide a way to set the search level in case the algorithm misses points like this.

fverdugo (changes requested):
@oriolcg Can you fabricate a simple example and add it that reproduces the bug that this PR wants to fix?

oriolcg:
@fverdugo, I added the test in https://github.com/gridap/Gridap.jl/pull/729/commits/6838ed5c5becc8e4d0785cb04819066cc108b637. The PR is ready

fverdugo (approved):

Issue #711, opened 2021-11-12
Hessian ad multifield


tmigot:
Thank you @amartinhuertas and @fverdugo for fixing the issue with the MultiField autodifferentiation in #687 . I think there is also the `hessian` function that needs to be fixed. For instance, the following script to get the hessian contribution was working with 0.15.5 but no longer. I tried combining gradient and jacobian, but this also leads too an error. Any idea how to recover this? Thanks!

```
using Gridap # works with Gridap 0.15.5, but not Gridap 0.17.5

n = 3
domain = (0, 1)

x0 = zeros(2)
xf = π * ones(2)
a = 1
c = 3
  
model = CartesianDiscreteModel(domain, n)
trian = Triangulation(model)
degree = 1
dΩ = Measure(trian, degree)
  
function f(x)
  φ, θ = x
  ∫(a * a * ∇(φ) ⊙ ∇(φ) + (c + a * (cos ∘ φ)) * (c + a * (cos ∘ φ)) * ∇(θ) ⊙ ∇(θ))dΩ
end
  
labels = get_face_labeling(model)
add_tag_from_tags!(labels, "diri1", [2])
add_tag_from_tags!(labels, "diri0", [1])
  
valuetype = Float64
reffe = ReferenceFE(lagrangian, valuetype, 1)
V0 = TestFESpace(
  model,
  reffe;
  conformity = :H1,
  labels = labels,
  dirichlet_tags = ["diri0", "diri1"],
)
V1 = TestFESpace(
  model,
  reffe;
  conformity = :H1,
  labels = labels,
  dirichlet_tags = ["diri0", "diri1"],
)
  
U0 = TrialFESpace(V0, [x0[1], xf[1]])
U1 = TrialFESpace(V0, [x0[2], xf[2]])
V = MultiFieldFESpace([V0, V1])
U = MultiFieldFESpace([U0, U1])
nU0 = Gridap.FESpaces.num_free_dofs(U0)
nU1 = Gridap.FESpaces.num_free_dofs(U1)
xin = zeros(nU0 + nU1)

yu = FEFunction(U, xin)
int = f(yu)
hess = Gridap.FESpaces._hessian(f, yu, int)
```

fverdugo:
I think that hessian can be implemented following the same lines as the Jacobian. See this file: https://github.com/gridap/Gridap.jl/blob/master/src/MultiField/MultiFieldFEAutodiff.jl

tmigot:
Closed via #731

PR #731, opened 2021-12-10
Ad multifield hessian function

tmigot:
Implement hessian AD for multifield following #687 
This should solve #711


fverdugo:
@tmigot Thanks for this PR!

Issue #709, opened 2021-11-10
Newest Vertex Bisection


aerappa:
This is a feature request for the [newest vertex bisection algorithm](https://en.wikipedia.org/wiki/Newest_vertex_bisection). For a very nice overview of the algorithm including MATLAB code see [this paper by Long Chen](https://www.math.uci.edu/~chenlong/Papers/Chen.L2006a.pdf).

Based on preliminary discussion, the interface for generating one level of refinement might be something like

`model_ref = newest_vertex_bisection(model, marked; hanging_allowed=false)`

where `model` is the current discrete model, `marked` is a list of cells to refine and `hanging_allowed` is a flag to specify if conformity must be preserved.

@fverdugo has kindly provided a skeleton version of how this might be achieved:

```julia
model = # whatherver discrete model you want
grid = get_grid(model)
node_coords = get_node_coordinates(grid)
cell_node_ids  = get_cell_dof_ids(grid)
reffes = get_reffes(grid)

# Now compute your "refined" quantities
nodal_coords_ref = 
cell_node_ids_ref =

# Build the refined grid
ncells_ref = length(cell_node_ids_ref)
cell_type_ref = fill(1,ncells_ref )
 grid_ref = UnstructuredGrid(nodal_coords_ref,cell_node_ids_ref,reffes,cell_type_ref)

# Build the model
topo_ref = GridTopology(grid_ref)
labels_ref = FaceLabeling(topo_ref)
model_ref = DiscreteModel(grid_ref,topo_ref,labels_ref)
```

This would ideally fit in as the last step in the standard

SOLVE ->  ESTIMATE -> MARK -> **REFINE**

loop. Thus, `FEFunction`s would need to be interpolated (hopefully in a rather lightweight way) between the various mesh levels of the refinement process. My initial thought was that this may be achievable using `BackgroundTriangulation`s since the algorithm produces a hierarchical set of Triangulations i.e. T_k​⊂T_{k+1} for all k.​

fverdugo:
Hi @aerappa!

The low-level functionality is clear. Now, we need to think a nice high-level API to express the solve/estimate/mark/refine loop in such a way that we can consider several different refinement strategies. Let me think a bit a bout this.

fverdugo:
Hi @aerappa 

as a first step, we can implement a simpler version of the problem and add more complexity later. I think that it can also be a good exercise to learn more Gridap details.

So for the moment I would do:

1. Just refinement of a grid
[...]

PR #733, opened 2021-12-16
Newest vertex bisection in 2D [WIP]

aerappa:
Addresses #709. All tests passing on my machine.

A couple of sticking points that I had:

1. `get_cell_node_ids(grid)` returns a Gridap custom type `Table`. It seems this is not intended to be mutable. However, the algorithm I am using requires changing rows for node_ids. For now I a copying this object to a `Matrix` and doing operations on the matrix. However, I also have another branch where I tried to stick with the table until I got stuck on the mutability. This is probably pretty bad for the memory footprint as well.
2. As I mentioned in the issue, I am sort of "cheating" by not recomputing the labels manually as @fverdugo suggested. Perhaps we can now discuss what he had in mind by this.
3. I only added two files, one in `/src` and one in `/test`. I'm not sure exactly what organizational structure is desired.


fverdugo:
Hi @aerappa is this ready to review?

aerappa:
Hi @fverdugo yes, in the sense that I am no longer adding commits. However, I have some questions (what I called sticking points 1, 2, 3 above) that may change the code as it stands. I would not want you to waste time reading the code in detail if the answers to these questions result in modifications. So maybe it's better to discuss them first?

fverdugo:
> get_cell_node_ids(grid) returns a Gridap custom type Table. It seems this is not intended to be mutable. However, the algorithm I am using requires changing rows for node_ids. For now I a copying this object to a Matrix and doing operations on the matrix. However, I also have another branch where I tried to stick with the table until I got stuck on the mutability. This is probably pretty bad for the memory footprint as well.

We also need to mutate `Table`s. To this end, we access the fields `.data` and `.ptrs` from the `Table` object directly . They are such that `a[i][j] == a.data[ a.ptrs[i]-1+j ]` for `a::Table`. This is a pretty standard way of storing a vector of vectors with 2 consecutive chunks of memory. It it also related with sparse matrix CSR/CSC storage. You are provably already familiar with this.

If you can guarantee  that `t` is a `Table` in a certain code location, feel free to access `t.data` and `t.ptrs` and mutate them (they are plain Julia vectors) as needed. We do this in different places of the code. If you cannot guarantee that `z` is a Table then convert it to a table: `t = Table(z)` and work (and possibly return) `t`. Note that this conversion does nothing if `z` happens to already be a Table.

src/Geometry/RefinedDiscreteModels.jl
```diff
@@ -0,0 +1,298 @@
...
+  sort_flag = false,
+)
+  # Not sure if necessary to keep old model unchanged. For my tests I use this
+  model_c = deepcopy(model)
+  grid = get_grid(model_c)
+  top = get_grid_topology(model_c)
+  ref_grid = newest_vertex_bisection(grid, top, η_arr, θ, sort_flag)
+  ref_topo = GridTopology(ref_grid)
+  #ref_labels = # Compute them from the original labels (This is perhaps the most tedious part)
+  ref_labels = FaceLabeling(ref_topo)
```
fverdugo, line 275: As you mentioned, this forgets about the old labels. Ideally, one wants to refine the given labels instead of creating new unrelated ones.

test/runtests.jl
```diff
@@ -2,6 +2,9 @@ module GridapRunTests
 
 using Test
 
+# TODO: temporary for testing
+@time @testset "Refinement" begin include("GeometryTests/RefinedDiscreteModelsTests.jl") end
```
fverdugo, line 6: I would call the file NewestVertexBisectionTests.jl

src/Geometry/RefinedDiscreteModels.jl
```diff
@@ -0,0 +1,298 @@
...
+
+# step 2
+function newest_vertex_bisection(
+  model::DiscreteModel,
+  η_arr::AbstractArray{<:AbstractFloat};
+  θ = 1.0, # corresponds to uniform refinement
+  sort_flag = false,
+)
+  # Not sure if necessary to keep old model unchanged. For my tests I use this
+  model_c = deepcopy(model)
```
fverdugo, line 269: The deepcopy here is quite memory consuming. I would avoid it by trying to use `model` read-only. If you need to mutate some data of the input model for some reason, selectively copy this data.

src/Geometry/RefinedDiscreteModels.jl
```diff
@@ -0,0 +1,298 @@
...
+  node_coords = get_node_coordinates(grid)
+  cell_node_ids = get_cell_node_ids(grid)
+  if sort_flag
+    cell_node_ids_ccw = sort_cell_node_ids_ccw(cell_node_ids, node_coords)
+  else
+    cell_node_ids_ccw = vcat(cell_node_ids'...)
+  end
+  node_coords_ref, cell_node_ids_ref =
+    newest_vertex_bisection(top, node_coords, cell_node_ids_ccw, η_arr, θ, sort_flag)
+  # TODO: Should not convert to matrix and back to Table
```
fverdugo, line 251: `cell_node_ids_ref` can be built as a Table from the beginning.

src/Geometry/RefinedDiscreteModels.jl
```diff
@@ -0,0 +1,298 @@
...
+  reffes = get_reffes(grid)
+  cell_types = get_cell_type(grid)
+  # TODO : Gracefully handle cell_types
+  new_cell_types = fill(1, length(cell_node_ids_ref) - length(cell_node_ids))
+  cell_types = [cell_types; new_cell_types]
+  UnstructuredGrid(node_coords_ref, cell_node_ids_ref, reffes, cell_types)
+end
+
+# step 2
+function newest_vertex_bisection(
```
fverdugo, line 262: Add a doc-string explaining what the function does and the meaning of each input and output.

fverdugo (changes requested):
Hi @aerappa Thanks for your work put into this PR! Please find a first round of comments.

src/Geometry/RefinedDiscreteModels.jl
aerappa: >We also need to mutate Tables. To this end, we access the fields .data and .ptrs from the Table object directly . They are such that a[i][j] == a.data[ a.ptrs[i]-1+j ] for a::Table. This is a pretty standard way of storing a vector of vectors with 2 consecutive chunks of memory. It it also related with sparse matrix CSR/CSC storage. You are provably already familiar with this.

Yes, I am familiar with this storage format. This works great! I am able to modify the rows of a `Table` now. However, I am still seeing a large memory footprint when I add a new row to a `Table` using `append_tables_globally` via something like

```julia
new_row = #some vector
elem = append_tables_globally(elem, Table([new_row])) 
```
is there a better way to achieve this? I saw some `copy`s in the definition of `append_tables_globally` for this operation. But maybe it's unavoidable since we don't know the memory size a priori. In this case though, I do know beforehand that I am just appending a new row though..

fverdugo: > But maybe it's unavoidable since we don't know the memory size a priori.

If you don't know the memory size a priori, then perhaps the best option is to work with a vector of vectors and use push! And finally convert this to a Table just before building the UnstructuredGrid. Function append_tables_globally is clearly not designed to be called at each cell.

fverdugo:
@aerappa let me know when you are ready for a second review round.

aerappa:
Hi @fverdugo , I believe I have made all your suggested changes, so please feel free to have another pass over the code.

However, I have been doing some testing to try to get an AMR scheme to work for a proof of concept. My estimator requires the construction of local Raviart-Thomas spaces. Based on my empirical tests, it seems that the Raviart-Thomas spaces indeed do rely on `cell_node_ids = get_cell_node_ids(grid)` being sorted by global ID. At least, when I sort `cell_node_ids` before constructing the new `DiscreteModel`, I can build the RT spaces, but if I do not I get an error deep inside Gridap. I can attach the stacktrace if that's useful.

 This is problematic, because the current algorithm requires a new ordering where the so called "peak" node- the node opposite the side to be bisected next- to be the first index in these arrays. If one wants to do multiple levels of refinement, this information needs to be conserved  somewhere.

A couple of potential solutions I could think of:

1. Create an auxiliary array that stores the indices of the peak nodes and only work with the sorted version of `cell_node_ids`
1. In the RT code, always sort by GID to ensure the ordering needed

I imagine the first option would be more attractive than the second but I would like to hear your opinion.

 Perhaps we can have a short discussion at some point? There is another "issue" that I highlighted with TODOs that `LazyArray`s do not seem to support `resize!` so I need to do a copies into standard arrays to do the dynamic modifications necessary for adaptive refinement.

test/runtests.jl
```diff
@@ -2,6 +2,9 @@ module GridapRunTests
 
 using Test
 
+# TODO: temporary for testing
+@time @testset "Refinement" begin include("GeometryTests/NewestVertexBisectionTests.jl") end
```
fverdugo, line 6: Move this to `test/GeometryTests/runtests.jl` and remove the TODO line.

src/Geometry/NewestVertexBisection.jl
```diff
@@ -0,0 +1,308 @@
...
+  issetequal_bitvec = issetequal(face_vec, face_top)
+  all(issetequal_bitvec)
+end
+
+function _get_midpoint(ngon::AbstractArray{<:VectorValue})
+  sum(ngon) / length(ngon)
+end
+
+function Base.atan(v::VectorValue{2,T}) where {T<:AbstractFloat}
+  atan(v[2], v[1])
```
fverdugo, line 175: Do not add this function (See bellow).

BTW, is this correct, v[2],v[1]?  Perhaps v[1],v[2] ?

src/Geometry/NewestVertexBisection.jl
```diff
@@ -0,0 +1,308 @@
...
+end
+
+function Base.:^(v::VectorValue{N,T}, r::Integer) where {T, N}
```
fverdugo, line 178: Do not add this function (see above).

src/Geometry/NewestVertexBisection.jl
```diff
@@ -0,0 +1,308 @@
...
+    elem::AbstractVector{<:AbstractVector{T}},
+    node::AbstractVector{<:VectorValue},
+    NT::T,
+) where {T<:Integer}
+    edgelength = zeros(NT, 3)
+    for t = 1:NT
+      elem_t = elem[t][:]
+        for (j, e) in enumerate(elem_t)
+            arr = filter(x -> x != e, elem_t)
+            diff = sqrt(sum((node[arr[1]] - node[arr[2]]) .^ 2))
```
fverdugo, line 23: Replace with
```julia
diff = norm(node[arr[1]] - node[arr[2]]) 
```

src/Geometry/NewestVertexBisection.jl
```diff
@@ -0,0 +1,308 @@
...
+LONG CHEN (2008). SHORT IMPLEMENTATION OF BISECTION IN MATLAB.
+In Recent Advances in Computational Sciences. WORLD SCIENTIFIC.
+DOI: 10.1142/9789812792389_0020
+
+# Arguments
+
+ -`model::DiscreteModel`: The current DiscreteModel to be refined.
+
+ -`η_arr::AbstractArray`: The values of the estimator on each cell of the domain
+   ordered in a consistent way with Geometry.Gridap.get_faces.
```
fverdugo, line 285: Which is the length of this array? num_cells(model) ? If yes, mention this instead of the relation with get_faces.

src/Geometry/NewestVertexBisection.jl
```diff
@@ -0,0 +1,308 @@
...
+)
+  reset_timer!(to)
+  disable_timer!(to)
+  # Not sure if necessary to keep old model unchanged. For my tests I use this
+  grid = get_grid(model)
+  top = get_grid_topology(model)
+  ref_grid = newest_vertex_bisection(grid, top, η_arr, θ, sort_flag)
+  ref_topo = GridTopology(ref_grid)
+  #ref_labels = # Compute them from the original labels (This is perhaps the most tedious part)
+  ref_labels = FaceLabeling(ref_topo)
```
fverdugo, line 306: As we mention, this does not take into account the flags of the input model. We can discuss how to solve this.

src/Geometry/NewestVertexBisection.jl
```diff
@@ -0,0 +1,308 @@
...
+end
+
+function Base.:^(v::VectorValue{N,T}, r::Integer) where {T, N}
+  VectorValue([v[i]^r for i = 1:N]...)
+end
+
+function _sort_ccw(cell_coords::AbstractVector{<:VectorValue})
+  midpoint = _get_midpoint(cell_coords)
+  offset_coords = cell_coords .- midpoint
+  sortperm(offset_coords, by = atan)
```
fverdugo, line 185: Replace `by = atan` with `by = v -> atan(v[2], v[1])`

src/Geometry/NewestVertexBisection.jl
```diff
@@ -0,0 +1,308 @@
+# using Gridap.Arrays
+using SparseArrays
+using TimerOutputs
+using Random
+
+# Create a TmerOutput, this is the main type that keeps track of everything.
+const to = TimerOutput()
```
fverdugo, line 7: We still need to decide which timing mechanism adopt for internal functions in Gridap.

For the moment, I would remove the timings until we have decided how to do this for other parts of Gridap. Hope it is not a big inconvenience for you.

Another temporary option is to remove the global constant `to` and pass it to the `newest_vertex_bisection` as an optional argument.

src/Geometry/NewestVertexBisection.jl
```diff
@@ -0,0 +1,308 @@
...
+  for k = 1:NE
+    i = edge[k, 1]
+    j = edge[k, 2]
+    d2p[i, j] = k
+    d2p[j, i] = k
+  end
+  d2p
+end
+
+function is_against_top(face::Table{<:Integer}, top::GridTopology, d::Integer)
```
fverdugo, line 156: Rename it to _is_against_top

src/Geometry/NewestVertexBisection.jl
```diff
@@ -0,0 +1,308 @@
...
+  face_vec = sort.([face[i][:] for i = 1:size(face, 1)])
+  face_top = sort.(get_faces(top, d, 0))
+  issetequal_bitvec = issetequal(face_vec, face_top)
+  all(issetequal_bitvec)
+end
+
+function is_against_top(face::Matrix{<:Integer}, top::GridTopology, d::Integer)
```
fverdugo, line 163: Idem.

fverdugo (changes requested):
Thanks @aerappa for the changes. See some more comments.

src/Geometry/NewestVertexBisection.jl
aerappa: @fverdugo I was going off of [this definition](https://docs.julialang.org/en/v1/base/math/#Base.atan-Tuple%7BNumber%7D) with y = v[2], x = v[1] is that correct to assume?

PR #734, opened 2021-12-29
Bugfix in `reference_grid(TET)`

pmartorell:
Dear @fverdugo ,
this an small bugfix on the changes introduced in #704 

The new tests also checks that only the boundary facets have one single cell around.


fverdugo:
@pmartorell Thanks for the fix. Have a nice new year's eve!

PR #736, opened 2022-01-19
Pvd functions

oriolcg:
Gridap functions that wrap `paraview_collection` to allow a common API  with `GridapDistributed`

### DIFF
--- a/src/Visualization/Vtk.jl
+++ b/src/Visualization/Vtk.jl
@@ -27,22 +27,8 @@ function createpvd(parts::Nothing,args...;kwargs...)
   createpvd(args...;kwargs...)
 end
 
-function createpvd(f::Function,args...;kwargs...)
-  pvd = createpvd(args...;kwargs...)
-  try
-    f(pvd)
-  finally
-    savepvd(pvd)
-  end
-end
-
-function createpvd(f::Function,parts::Nothing,args...;kwargs...)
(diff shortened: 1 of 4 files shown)


src/Visualization/Vtk.jl
```diff
@@ -17,6 +17,40 @@ function createvtk(args...;kwargs...)
...
+function createpvd(f::Function,args...;kwargs...)
+  pvd = createpvd(args...;kwargs...)
+  try
+    f(pvd)
+  finally
+    savepvd(pvd)
+  end
+end
+
+function createpvd(f::Function,parts::Nothing,args...;kwargs...)
```
fverdugo, line 39: I think that this one is redundant. Previous function already does the job for this case, right?

fverdugo (changes requested):
Hi @oriolcg thanks for this! See my minor comment.

src/Visualization/Vtk.jl
fverdugo: Even better: keep this function but remove `::Function` so that one cas use other types of callable objects.

fverdugo (approved):

src/Visualization/Vtk.jl
fverdugo: I have quick-fixed this myself.

fverdugo (approved):

PR #740, opened 2022-02-07
Bug-fixed _flatten_upper_triangle generated function

amartinhuertas:
Aslo added a test that reproduces the BUG before the fix


fverdugo (approved):
Very good point @amartinhuertas !

PR #751, opened 2022-02-28
fill! to fillstored!

oriolcg:
Replacing `fill!` by `fillstored` in GridapODEs functions. This PR needs to be released for PR gridap/GridapDistributed.jl#81 to pass the tests.


oriolcg:
Leaving `fill!` for dense matrices since `fillstored!` is not implemented for this case. Using only `fillstored!` when `issparse(A)` returns  `true`.

santiagobadia:
@oriolcg can we create a method that does the dispatch for each case?
I think it would be better than replicating the if statements in so many places

oriolcg:
> The other question is, do we really need to run the code for dense matrices?

> If not, I would use fillstored! everywhere

We were using dense matrices in the tests. I defined sparse matrix in `ODEOperatorMocks` and left `fillstored!` only in https://github.com/gridap/Gridap.jl/pull/751/commits/a02502870aecb5ce2840bffe73a516ed906e5e06.

santiagobadia:
Ok much better. Thanks!

santiagobadia:
Tj6btbbhkhb

oriolcg:
Hi @santiagobadia , this PR has been closed without being merged.

Issue #747, opened 2022-02-25
Interpolate funs at arbitrary points inside _active_ domain?


ericneiva:
Hi @fverdugo, @oriolcg, @Balaje,

I want to interpolate a FE function defined in an _active_ triangulation (i.e., a strict subset of the background triangulation) at a set of points inside the active domain.

I think the machinery finds the points in the cells correctly, but the cells are indexed according to the background triangulation :point_right: https://github.com/gridap/Gridap.jl/blob/154ee18e27ae1661db99843f6e808c805fffbb54/src/CellData/CellFields.jl#L829 This leads to some out of bounds errors.

Do you think it makes sense to have `get_active_model` instead of `get_background_model` in the line above? If the FE function is only defined in the active triangulation, then we cannot evaluate it outside the active domain... (Of course, with `get_active_model` I get the outcome I am looking after.)

oriolcg:
To me it make sense. Does `get_active_model` fall back to `get_background_model` in case no active triangulation has been specified? If that's the case, I don't see any reason for not to make the change.

ericneiva:
Yes, under `get_active_model` we restrict the model to the active triangulation. If we pass active = background then it devolves a trivial restriction.

Ok, then, if you agree, I'll prepare a branch with the changes and push.

PR #752, opened 2022-03-01
Restrict to active model the evaluation of a FE function at arbitrary points

ericneiva:
Please, see related discussion in #747.


oriolcg:
This PR looks good to me.

PR #755, opened 2022-03-03
Implement addition and subtraction between TensorValue and SymTensorValue

ericneiva:
Implements missing addition and subtraction between `TensorValue` and `SymTensorValue` types (+ tests).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- Implemented addition/subtraction between a `TensorValue` and a `SymTensorValue`. Since PR [#755](https://github.com/gridap/Gridap.jl/pull/755).
+
 ### Fixed
 - Restrict to the active model the evaluation of a FE function at arbitrary points. Since PR [#752](https://github.com/gridap/Gridap.jl/pull/752).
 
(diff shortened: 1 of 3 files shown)


amartinhuertas:
Hi @ericneiva ! Do you plan to use symmetric tensors in gridap ? I have explored their usage in gridaphybrid and detected several bugs that I have fixed there temporarily but I might expedite fixing them in gridap if you are going to need them.

ericneiva:
Hi, @amartinhuertas!

Yes, I am already using them on a driver which solves the linear elasticity problem on a 2-manifold. It requires subtracting symmetric and non-symmetric tensors. After this PR, I can integrate the weak form and everything appears to work.

I just have one question: Do any of these bugs provoke that your hybrid methods do not converge at the expected rate? If not, I'd say I do not depend on your fixes in the short-medium term.

Thanks a lot for your message :slightly_smiling_face:

amartinhuertas:
There are a number of bugs. E.g. interpolation of analytical function inside Fe space

Here are the fixes fyi 

https://github.com/gridap/GridapHybrid.jl/blob/main/src/tmp_fixes_symmetric_valued_lagrangian_reffes.jl

ericneiva:
It appears those fixes do not affect my results... for now. 

I'll take a closer look later to double-check.

Thanks again, @amartinhuertas!

PR #756, opened 2022-03-03
Renaming GridapODEs to ODEs

oriolcg:
The name of `GridapODEs` module is redundant, changing it to `ODEs`.


fverdugo:
If you really want to be consistent with semantic versioning, you can also temporarily add `const GridapODEs = ODEs` rigth after closing the `ODEs` module so that old code using `Gridap.GridapODEs` still works (this will prevent to start gridap 0.18 just for this small fix). Of course this should be removed in the next minor release.

fverdugo (approved):
I find this renaming very convenient @oriolcg. 

Perhaps you want to take into account my previous comment about semantic versioning. But this is not very crucial at this moment since perhaps you are the only one already using the odes part from Gridap directly.

fverdugo:
BTW, this PR seems to have changes from PR #751. I would merge #751 before this one.

oriolcg:
Done in 0a4ffb5d4657cce31104446ca657a9d806160997. I accidentally pushed a version with PR #751 locally merged. It would be good to first accept PR #751 and then accept this PR.

fverdugo:
> Done in https://github.com/gridap/Gridap.jl/commit/0a4ffb5d4657cce31104446ca657a9d806160997. I accidentally pushed a version with PR https://github.com/gridap/Gridap.jl/pull/751 locally merged. It would be good to first accept PR https://github.com/gridap/Gridap.jl/pull/751 and then accept this PR.

We wrote at the same time !

santiagobadia:
@oriolcg can you solve the conflicts?

oriolcg:
> @oriolcg can you solve the conflicts?

Done in https://github.com/gridap/Gridap.jl/pull/756/commits/b93eaafe4d7828d5e95cd0a8b6af4e0e33436a2a

PR #766, opened 2022-03-14
Fixed issue 762, writing vtk files with empty triangulations

principejavier:


fverdugo:
Thanks @principejavier! I am going to add some tests.

PR #777, opened 2022-04-25
Add modal c0 machinery

ericneiva:
Hi, @santiagobadia,

This PR adds (generalised) Modal C0 bases and reffes into Gridap.

I have refactored the code such that it is minimally invasive.

There are two things however to improve:

1. Modal C0 reffes are generally different at each cell, because they express different shape functions according to the cell bounding box. In the implementation, I am assigning a different Modal C0 cell type to each cell. On the other hand, the cell-wise ownership given by `CellConformity` is compressed on cell types. At the moment, I compute the `CellConformity` for each Modal C0 cell type. This is redundant. We have different shape functions, but the `CellConformity` is usually the same everywhere. I have found out that this approach is quite inefficient in https://github.com/gridap/Gridap.jl/blob/1fb3dc9abf8c47685637901bd14a74e4355a9492/src/FESpaces/ConformingFESpaces.jl#L78 and https://github.com/gridap/Gridap.jl/blob/1fb3dc9abf8c47685637901bd14a74e4355a9492/src/FESpaces/ConformingFESpaces.jl#L40 We should definitely compute once (the shared) `CellConformity` and assign it to all Modal C0 cell types. 
2. In order to reuse the procedure that creates the AgFE space on top of the standard one, I am expressing the Modal C0 shape functions as a trivial linear combination in https://github.com/gridap/Gridap.jl/blob/eadad319241a0163b31381153d9b0c877d03a127/src/ReferenceFEs/ReferenceFEInterfaces.jl#L491 If I don't do this I cannot evaluate dofs at root shape functions in the physical domain, as `Gridap` errors at attempting to evaluate a single Modal C0 shapefun at a single point. I think `Gridap`'s design is planned to evaluate the whole Modal C0 basis at a single point, not a single shapefun. Maybe @fverdugo knows what is missing.

We can discuss about those two points whenever you want. 

In the meantime, the new functionality (up to the points above) can be reviewed and merged into the code.

Cheers!


fverdugo:
> I think Gridap's design is planned to evaluate the whole Modal C0 basis at a single point, not a single shapefun. Maybe @fverdugo knows what is missing.

It is  possible (at least in theory) to evaluate a single function at a single point. But this situations is not used often, and thus it is likely that there is missing implementations related with this. If you provide a minimal example that crashes with this error, I can evaluate what is still missing.

ericneiva:
Hi, @fverdugo,

If you run the 1D snippet below checking out on the previous commit of this branch, i.e., 1bf6f6c6, you should get the error I am referring to:

```julia
using Test
using Gridap
using Gridap.CellData
using Gridap.FESpaces

u(x) = x[1]^3

domain = (0,4)
partition = (4,)
model = CartesianDiscreteModel(domain,partition)

order = 3
reffe = ReferenceFE(modalC0,Float64,order)
V = FESpace(model,reffe,conformity=:H1)

shfns_g = get_fe_basis(V)
phys_shfns_g = change_domain(shfns_g,PhysicalDomain())
dofs_f = get_fe_dof_basis(V)
dofs_f(phys_shfns_g)
```

Of course, it works running on the newest commit of this branch, where the only diff is that I apply the trick of expressing the Modal C0 shapefuns as a trivial linear combination.

Thanks for taking a look at this.

fverdugo (approved):
After talking with @ericneiva, the linear combination trick looks good to me.
Performance issues can be fixed in a second round.

ericneiva:
@santiagobadia If you agree and there are no additional comments from your side, could you please accept the PR?

PR #797, opened 2022-06-10
Towards AD for Skeleton integration terms - gradient added

kishore-nori:
* Added functionality for performing gradient of functionals involving integrals over Skeleton faces
* corresponding tests for the new constructs and functionalities have also been added
* Jacobian and Hessian of functionals involving Skeleton integration terms is work-in-progress

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+- Added functionality to take gradient of functional involving integration (`DomainContribution`) over Skeleton faces, with respect to the degrees of freedom `FEFunction`.  
+
 ## [0.17.13] - 2022-05-31
 
 ### Added
--- a/src/CellData/SkeletonCellFieldPair.jl
+++ b/src/CellData/SkeletonCellFieldPair.jl
@@ -1,0 +1,98 @@
+"""
+  SkeletonCellFieldPair is a special construct for allowing uh.plus
+  and uh.minus to be two different CellFields. In particular, it is
+  useful when we need one of the CellFields to be the dualized
+  version of the other for performing ForwardDiff AD of a Skeleton
+  integration DomainContribution wrt to the degrees of freedom
+  of the CellField, plus and minus sensitivities done separately,
+  so as to restrict the interaction between the dual numbers.
+
+  It takes in two CellFields and stores plus version of CellFieldAt
+  for the first CellField and minus version of CellFieldAt for the
+  second the CellField. SkeletonCellFieldPair is associated with the
+  same triangulations as the CellFields
+"""
+struct SkeletonCellFieldPair{
+  P<:CellField, M<:CellField, T<:Triangulation} <: CellField
+
+  cf_plus::P
+  cf_minus::M
+  trian::T # Background Triangulation to which CellFields belong to
+
+  function SkeletonCellFieldPair(
+    cf_plus_plus::CellFieldAt,
+    cf_minus_minus::CellFieldAt,
+    trian::Triangulation)
+
+    P = typeof(cf_plus_plus)
+    M = typeof(cf_minus_minus)
+    T = typeof(trian)
+
+    new{P,M,T}(cf_plus_plus,cf_minus_minus,trian)
+  end
+end
+
+function SkeletonCellFieldPair(
+  cf_plus::CellField,
+  cf_minus::CellField,
+  strian::SkeletonTriangulation)
+
+  @check DomainStyle(cf_plus) == DomainStyle(cf_minus)
+  cf_plus_trian = get_triangulation(cf_plus)
+  cf_minus_trian =  get_triangulation(cf_plus)
+  @notimplementedif !(cf_plus_trian ===  cf_minus_trian)
+
+  @check num_dims(cf_plus_trian) == num_dims(strian) + 1
+  @check get_background_model(cf_plus_trian) === get_background_model(strian)
+
+  cf_plus_plus  = cf_plus.plus
+  cf_minus_minus = cf_minus.minus
+
+  SkeletonCellFieldPair(cf_plus_plus,cf_minus_minus,cf_plus_trian)
+end
+
+function Base.getproperty(a::SkeletonCellFieldPair,sym::Symbol)
+  if sym in (:⁺,:plus)
+    getfield(a,:cf_plus)
+  elseif sym in (:⁻, :minus)
+    getfield(a,:cf_minus)
+  else
+    getfield(a,sym)
+  end
+end
+
+function DomainStyle(a::SkeletonCellFieldPair)
+  DomainStyle(getfield(a,:cf_plus))
+end
+
+function get_triangulation(a::SkeletonCellFieldPair)
+  a.trian
+end
+
+function change_domain(a::SkeletonCellFieldPair,target_trian::Triangulation,target_domain::DomainStyle)
+  change_domain(a.plus,DomainStyle(a),target_trian,target_domain)
+end
+
+function Base.propertynames(a::SkeletonCellFieldPair, private::Bool=false)
+  (fieldnames(typeof(a))...,:⁺,:plus,:⁻,:minus)
+end
+
+function Base.show(io::IO,::MIME"text/plain",f::SkeletonCellFieldPair)
+  show(io,f)
+end
+
+# derivatives of the SkeletonCellFieldPair
+
+function gradient(a::SkeletonCellFieldPair)
+  grad_cf_plus_plus = gradient(getfield(a,:cf_plus))
+  grad_cf_minus_minus = gradient(getfield(a,:cf_minus))
+  SkeletonCellFieldPair(
+    grad_cf_plus_plus,grad_cf_minus_minus,getfield(a,:trian))
+end
+
+function ∇∇(a::SkeletonCellFieldPair)
+  hess_cf_plus_plus = ∇∇(getfield(a,:cf_plus))
+  hess_cf_minus_minus = ∇∇(getfield(a,:cf_minus))
+  SkeletonCellFieldPair(
+    hess_cf_plus_plus,hess_cf_minus_minus,getfield(a,:trian))
+end
--- a/src/Arrays/Autodiff.jl
+++ b/src/Arrays/Autodiff.jl
@@ -106,3 +106,49 @@ function evaluate!(result,k::AutoDiffMap,ydual,x,cfg::ForwardDiff.JacobianConfig
   ForwardDiff.extract_value!(T, result, ydual)
   return result
 end
+
+# overloads for AD of SkeletonTriangulation DomainContribution
+
+# earlier the dispatch of autodiff_array_gradient for Skeleton terms was
+# based on the SkeletonPair (which is intuitive as it has fields plus and minus)
+# autodiff_array_gradient(a,i_to_x,j_to_i::SkeletonPair)
+# but src/Arrays/ can import SkeletonPair for the above dispatch due to
+# circular dependency of modules
+# so now we just tuple instead of SkeletonPair, where the first entry of the
+# tuple corresponds to plus side and the second to minus
+
+function autodiff_array_gradient(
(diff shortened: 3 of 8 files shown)


amartinhuertas:
Hi @kishore-nori ! Thanks for the PR! 

The tests did not pass! https://github.com/gridap/Gridap.jl/runs/6827393445?check_suite_focus=true#step:6:211

Seems an easy fix.

NEWS.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+- Added functionality to take gradient of functional involving integration (`DomainContribution`) over Skeleton faces, with respect to the degrees of freedom `FEFunction`.
```
amartinhuertas, line 10: Add a pointer to the PR as in the other entries.

kishore-nori: sure Alberto will do it soon!

src/CellData/SkeletonCellFieldPair.jl
```diff
@@ -0,0 +1,98 @@
...
+function DomainStyle(a::SkeletonCellFieldPair)
+  DomainStyle(getfield(a,:cf_plus))
+end
+
+function get_triangulation(a::SkeletonCellFieldPair)
+  a.trian
+end
+
+function change_domain(a::SkeletonCellFieldPair,target_trian::Triangulation,target_domain::DomainStyle)
+  change_domain(a.plus,DomainStyle(a),target_trian,target_domain)
```
amartinhuertas, line 73: Put here a comment where you argue that we select a.plus arbitrarily, while justifying why we can afford that.

src/CellData/SkeletonCellFieldPair.jl
```diff
@@ -0,0 +1,98 @@
...
+  and uh.minus to be two different CellFields. In particular, it is
+  useful when we need one of the CellFields to be the dualized
+  version of the other for performing ForwardDiff AD of a Skeleton
+  integration DomainContribution wrt to the degrees of freedom
+  of the CellField, plus and minus sensitivities done separately,
+  so as to restrict the interaction between the dual numbers.
+
+  It takes in two CellFields and stores plus version of CellFieldAt
+  for the first CellField and minus version of CellFieldAt for the
+  second the CellField. SkeletonCellFieldPair is associated with the
```
amartinhuertas, line 12: Remove first "the"

amartinhuertas: Perhaps we may mention that we conceive SkeletonCellFieldPair as in internal code artifact of Gridap and we do not pretend the user to deal with it.

src/CellData/SkeletonCellFieldPair.jl
```diff
@@ -0,0 +1,98 @@
...
+  P<:CellField, M<:CellField, T<:Triangulation} <: CellField
+
+  cf_plus::P
+  cf_minus::M
+  trian::T # Background Triangulation to which CellFields belong to
+
+  function SkeletonCellFieldPair(
+    cf_plus_plus::CellFieldAt,
+    cf_minus_minus::CellFieldAt,
+    trian::Triangulation)
```
amartinhuertas, line 25: I would remove the `trian` argument from the inner constructor. In practice, we always extract it from either cf_plus_plus/ or cf_minus_minus.

src/CellData/SkeletonCellFieldPair.jl
```diff
@@ -0,0 +1,98 @@
...
+
+    P = typeof(cf_plus_plus)
+    M = typeof(cf_minus_minus)
+    T = typeof(trian)
+
+    new{P,M,T}(cf_plus_plus,cf_minus_minus,trian)
+  end
+end
+
+function SkeletonCellFieldPair(
```
amartinhuertas, line 35: Why do we actually need this outer constructor? I would remove it, while putting part of the checks inside it to the inner constructor.

amartinhuertas (changes requested):
@kishore-nori nice work! I have completed my first round of review. Please, take into account my minor comments and resubmit.

NEWS.md
kishore-nori: Done!

src/CellData/SkeletonCellFieldPair.jl
kishore-nori: regarding having `trian`  in the `SkeletonCellFieldPair`, i though it can provide quick access rather than go over `getproperty` multiple times.

src/CellData/SkeletonCellFieldPair.jl
kishore-nori: In the second constructor for `SkeletonCellFieldPair`  we want to check if `dim(train) == dim(strain) + 1` so I kept `strian`  as an argument in the constructor

src/CellData/SkeletonCellFieldPair.jl
amartinhuertas: Ok. We can keep it as a member variable of `SkeletonCellFieldPair`, but we do not need to have it as an argument to the constructor. This is dangerous as it opens a door to pass a `Triangulation` that has nothing to do with the one underlying the first two arguments.

kishore-nori:
Currently, there is a problem of requiring circular dependency due to `BlockMap` usage in `src/Arrays/Autodiff.jl`. The thing is, I think we ll have to move the `autodiff_array_gradient` (for Skeleton integration) to `FESpaces`, may be to `FEAutodiff.jl` or a different file as we are using `BlockMap` that is not imported (due ciruclar dependency) into `Gridap.Arrays`. `BlockMap`  belongs to Gridap.Fields which `FESpaces` is also using/importing.

The only changes to be done is to import `DualizeMap` and `AutoDiffMap` to `Gridap.FESpaces`. Based on your suggestions for the best way to go ahead, I ll make changes accordingly. This comes with an advantage of being able to `SkeletonPair` again!

src/CellData/SkeletonCellFieldPair.jl
kishore-nori: Done!

src/CellData/SkeletonCellFieldPair.jl
kishore-nori: Done!

NEWS.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+- Added functionality to take gradient of functional involving integration (`DomainContribution`) over Skeleton faces, with respect to the degrees of freedom `FEFunction`.  The interface remains the same - `gradient(fuh,uh)`. Since PR [#797](https://github.com/gridap/Gridap.jl/pull/797)
```
amartinhuertas, line 10: Should `gradient(fuh,uh)` actually be `gradient(uh)` ?

kishore-nori: Yes sorry it has to be `gradient(f,uh)`, will change it.

NEWS.md
amartinhuertas: Perfect.

NEWS.md
kishore-nori: Done!

src/CellData/SkeletonCellFieldPair.jl
amartinhuertas: Can we defer this check to a latter stage?

amartinhuertas:
> The only changes to be done is to import DualizeMap and AutoDiffMap to Gridap.FESpaces. Based on your suggestions for the best way to go ahead, I ll make changes accordingly. This comes with an advantage of being able to SkeletonPair again!

We can proceed as you suggest.

kishore-nori:
Sure Alberto, I ll move it `FESpaces`. Should I put it in `FEAutodiff.jl` or in a new file?

src/CellData/SkeletonCellFieldPair.jl
kishore-nori: yeah true, I ll make the changes and move the outer constructor as the inner constructor

src/CellData/SkeletonCellFieldPair.jl
kishore-nori: Yeah it can be done at a later stage when the integration on the Skeleton terms happens but this will be repeated every single time the integration happens over iterations, which is not necessary I guess.

amartinhuertas: > Yeah it can be done at a later stage when the integration on the Skeleton terms happens but this will be repeated every single time the integration happens over iterations, which is not necessary I guess.

I think it is better to defer this check when the integration on the Skeleton term happens. This is more aligned with how the Gridap high-level API is designed. Besides, there is no guarantee that the SkeletonTriangulation that you pass to the constructor is going to be the same as the one that you use in the weak formulation.

src/CellData/SkeletonCellFieldPair.jl
kishore-nori: Hi Alberto, made changes, please check..

amartinhuertas:
> Sure Alberto, I ll move it FESpaces. Should I put it in FEAutodiff.jl or in a new file?

you can put in FEAutodiff.jl along with a comment on why you had to put it there and not in Arrays.

kishore-nori:
Sure, I ll move it to `src/FESpaces/FEAutodiff.jl` and add details as suggest on why it has been placed there.

src/CellData/SkeletonCellFieldPair.jl
amartinhuertas: > Hi Alberto, made changes, please check..

see https://github.com/gridap/Gridap.jl/pull/797#discussion_r895264561

src/CellData/SkeletonCellFieldPair.jl
kishore-nori: ohk, so should I go ahead and remove the `strian` in the constructor and also remove the related check?

amartinhuertas: > ohk, so should I go ahead and remove the strian in the constructor and also remove the related check?

yes, please. The check should be actually moved to another place, not 100% removed from the project.

kishore-nori: yeah I ll do it after the CI completes, to see if there are more errors. But isn't the check, or some other compatibility check automatically happening when we integrate the `SkeletonCellFieldPair` over Skeleton Triangulation?

amartinhuertas: >  But isn't the check, or some other compatibility check automatically happening when we integrate the SkeletonCellFieldPair over Skeleton Triangulation?

Perhaps you are right. Can you try to find in the code where such a test is being performed?

kishore-nori: I ll try to dig into the code and check this and update..

amartinhuertas:
> Sorry this the latest commit. The thing is I need both the outer and inner constructors for convenience, the outer one for normal construction and inner one when I have derivatives of `SkeletonCellFieldPair`. The derivatives of `CellFieldAt` return a `CellFieldAt` so having the inner one would be useful.

ok.

amartinhuertas:
hi @kishore-nori ... FYI, it seems that Gridap tests got stuck in Github actions. I cancelled them after 2hours and 45 mins. 

https://github.com/gridap/Gridap.jl/runs/6856453767?check_suite_focus=true#step:6:210

I have re-run them to see if the problem persists.

amartinhuertas:
Can you run FESpaceTests.jl on your local machine? 

I am not sure what is going on, but it might be (hypothesis) that you have some sort of infinite recursion among the inner/outer constructions or something like that.

amartinhuertas:
Also which version of Julia do you have locally? The remote end uses julia version 1.7.3

kishore-nori:
Yes Alberto, I will test in on my laptop locally in sometime. I am currently having Julia 1.7.2 running.

kishore-nori:
Hi Alberto, you were right, there was recursive loop between inner and outer constructor of `SkeletonCellFieldPair`. I tested all the tests that have been affected by changes and they all run fine now, so pushed the changes and hopefully the CI will run fine now.

amartinhuertas (approved):

PR #799, opened 2022-06-16
Extended `gradient` for `MultiField` functionals involving Skeleton integration

kishore-nori:
* Refactored the existing `MultiField` functionals `gradient` code to make dispatch for Skeleton integration terms possible
* Extended the  `MultiField` functionals `gradient` for functionals involving Skeleton integration, in similar fashion as done for the `SingleField` case 
* Added tests to compare the results to that of `ForwardDiff`
* Added tests for `SkeletonCellFieldPair` in the context of `MultiField`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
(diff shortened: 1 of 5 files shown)


amartinhuertas (approved):
Hi @kishore-nori ! Thanks for the PR! Accepting it straight away.

PR #800, opened 2022-06-17
Changed the method for `SkeletonCellFieldPair` for AD to work with operations inside `mean` and `jump`

kishore-nori:
* Overloaded the `evaluate!` method for `SkeletonCellFieldPair`
* removed the `change_domain` call resulting in wrong behaviour with terms involving operations of `CellFields` inside `mean` and `jump` terms 
* added both `SingleField` and `MultiField` AD tests compared with `ForwardDiff` for such terms

### DIFF
--- a/src/CellData/SkeletonCellFieldPair.jl
+++ b/src/CellData/SkeletonCellFieldPair.jl
@@ -68,16 +68,16 @@ function get_triangulation(a::SkeletonCellFieldPair)
 end
 
 #=
-We arbitrarily choose plus side of SkeletonCellFieldPair for evaluations, as
-we don't use the DomainContributions associated with Ω and Γ while performing
-the sensitivities for SkeletonTriangulation (Λ) DomainContribution. This
-fix is for error free evaluation of the functional when a SkeletonCellFieldPair
-is passed into it. Ideally, if we could parse and extract only the Skeleton
-integration terms from the functional's julia function form, this fix is not
-required, but this is not trivial to do. On the positive side, since the
-evaluations are all lazy and not used, this doesn't put any noticable memory
-or computational overhead. Ofcourse, it is made sure that the such plus side
-pick doesn't happen when the integration over the SkeletonTriangulation
+We arbitrarily choose plus side of SkeletonCellFieldPair for evaluations, as we
+don't use the DomainContributions associated with Ω and Γ while performing the
+sensitivities for SkeletonTriangulation (Λ) DomainContribution. This fix is for
+error free evaluation of the functional when a SkeletonCellFieldPair is passed
+into it. Ideally, if we could parse and extract only the Skeleton integration
+terms from the functional's julia function form, this fix is not required, but
+this is not trivial to do. On the positive side, since the evaluations are all
+lazy and not used, this doesn't put any noticable memory or computational
+overhead. Ofcourse, it is made sure that the such plus side pick doesn't happen
+when the integration over the SkeletonTriangulation
 =#
 # If SkeletonCellFieldPair is evaluated we just pick the plus side parent
 function evaluate!(cache,f::SkeletonCellFieldPair,x::CellPoint)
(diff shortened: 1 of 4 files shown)


amartinhuertas:
Hi @kishore-nori ! Thanks for your work! Can you please edit NEWS.md?

kishore-nori:
> Hi @kishore-nori ! Thanks for your work! Can you please edit NEWS.md?

Done!

src/CellData/SkeletonCellFieldPair.jl
```diff
@@ -80,12 +76,28 @@ is passed into it. Ideally, if we could parse and extract only the Skeleton
...
-  change_domain(a.plus,DomainStyle(a),target_trian,target_domain)
+# If SkeletonCellFieldPair is evaluated we just pick the plus side parent
+function evaluate!(cache,f::SkeletonCellFieldPair,x::CellPoint)
+  _f, _x = _to_common_domain(f.plus.parent,x)
+  cell_field = get_data(_f)
+  cell_point = get_data(_x)
+  lazy_map(evaluate,cell_field,cell_point)
+end
+
+# fix for CellFieldAt{T}(parent::OperationCellField) giving right output
```
amartinhuertas, line 90: Can we try to improve this comment? What is the key idea we want to say here?

I guess we want to say that:
 1. `CellFieldAt{T}(OperationCellField)` results in calling `CellFieldAt{T}(operands)`.
 2. Without this override, `CellFieldAt{T}(::SkeletonCellField)` results in XXX, which is not correct.
 3. Thus, we need this override.

Correct?

kishore-nori: Yes exactly Alberto.

src/CellData/SkeletonCellFieldPair.jl
```diff
@@ -80,12 +76,28 @@ is passed into it. Ideally, if we could parse and extract only the Skeleton
...
+# or else it chooses the SkeletonCellFieldPair directly as the parent
+# resulting in errors and it not the intended behaviour to have the parent
+# as SkeletonCellFieldPair and is not consistent with our getproperty rules
+function CellFieldAt{T}(parent::SkeletonCellFieldPair) where T
+  getproperty(parent,T)
 end
 
+# to handle the evaluation of SkeletonCellFieldPair at BoundaryTriangulation
```
amartinhuertas, line 98: Ok. I do not see any problem in overriding `get_data`. Indeed, this is a method which belong to the abstract interface of a `CellField`. My only concern is ... are we 100% sure that this behaviour of `get_data` is the intended one in any of the calls to `get_data(::SkeletonCellFieldPair)`? (To think).

amartinhuertas (changes requested):
hi @kishore-nori ! Thanks for your work. We are almost there. I only made a pair of comments that you have to address.

src/CellData/SkeletonCellFieldPair.jl
kishore-nori: Yes true, I am was unsure about this fully. I checked for the cases of all trians if there could be a side effect, but there was wasn't.

kishore-nori:
> hi @kishore-nori ! Thanks for your work. We are almost there. I only made a pair of comments that you have to address.

Thank you for the review Alberto and the comments. I ll improve the comments and think more about the `get_data` method for `SkeletonCellFieldPair`, atleast from the tests on various triangulations, I don't see any side effects!

src/CellData/SkeletonCellFieldPair.jl
kishore-nori: The `get_data` is called in `change_domain_ref_ref` and `change_domain_phy_phs` when the `trian` for integrand and `quad` are not the same. This the case for `BoundaryTriangulation` and `SkeletonTriangulation`, choosing `plus` side is consistent with that of the choice we make in `evaluation!`. In case of integration over `SkeletonTriangulation` which hits inside `change_domain_ref_ref` specialised for `SkeletonTriangulation` the call `isa(get_array(a.cell_field),Fill)` gives error for `uh` and also `SkeletonCellFieldPair` so `get_data` doesn't effect here. In case of `BodyFittedTriangulation` there is no hit to `get_data` directly and `evaluate!` handles it as intended

kishore-nori:
> hi @kishore-nori ! Thanks for your work. We are almost there. I only made a pair of comments that you have to address.

Hi Alberto! I improved the comments and explain the choice of overloads, please check when you are free.

src/CellData/SkeletonCellFieldPair.jl
amartinhuertas: ok, so we can go ahead. Thanks for looking deeply into this.

amartinhuertas (approved):

PR #803, opened 2022-06-21
Extended the `jacobian` to functionals involving Skeleton terms

kishore-nori:
* `jacobian` for functionals involving inetgration over Skeleton terms can be handled now 
* Added corresponding tests compared with ForwardDiff
* The high-level API remains the same

### DIFF
--- a/src/FESpaces/FEAutodiff.jl
+++ b/src/FESpaces/FEAutodiff.jl
@@ -165,9 +165,7 @@ end
   Gridap, where we can use or import required functionalities from other
   modules without any circular dependency
 =#
-function autodiff_array_gradient(
-  a, i_to_x, j_to_i::SkeletonPair)
-
+function autodiff_array_gradient(a, i_to_x, j_to_i::SkeletonPair)
   i_to_xdual = lazy_map(DualizeMap(ForwardDiff.gradient),i_to_x)
 
   # dual output of both sides at once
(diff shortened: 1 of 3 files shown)


src/FESpaces/FEAutodiff.jl
```diff
@@ -191,12 +189,57 @@ function autodiff_array_gradient(
...
-  @notimplemented
+function autodiff_array_jacobian(a,i_to_x,j_to_i::SkeletonPair)
+  i_to_xdual = lazy_map(DualizeMap(ForwardDiff.jacobian),i_to_x)
+  j_to_ydual_plus, j_to_ydual_minus = a(i_to_xdual)
+
+  # bilinear form tested with test basis functions dv results in
+  # DomainContribution containing vector of vector blocks of 2 vectors
+  # coming from plus side and minus side of dv
+  # so we densify each of them into plain vectors and construct
+  # construct back the jacobian contributions into a MatrixBlock
```
amartinhuertas, line 200: construct twice

src/FESpaces/FEAutodiff.jl
```diff
@@ -191,12 +189,57 @@ function autodiff_array_gradient(
...
+  J_21 = lazy_map((x,b)->view(x, b+1:size(x,1),:),
+                  j_to_result_plus_dense,num_dofs_x_face_and_cell_minus)
+
+  J_12 = lazy_map((x,b)->view(x, 1:b,:),
+                  j_to_result_minus_dense,num_dofs_x_face_and_cell_plus)
+
+  J_22 = lazy_map((x,b)->view(x, b+1:size(x,1),:),
+                  j_to_result_minus_dense,num_dofs_x_face_and_cell_minus)
+
+  # Assembly on SkeletonTriangulation expects an array of interior of facets
```
amartinhuertas, line 236: Remove "of interior"

test/FESpacesTests/FEAutodiffTests.jl
```diff
@@ -162,4 +165,65 @@ gridapgradj = assemble_vector(gradient(j_Λ,uh),U)
...
+              jump(uh*n_Λ)⊙jump(vh*n_Λ))*dΛ
+g(uh,vh) = ∫(mean(uh*vh))*dΛ
+h(uh,vh) = ∫(jump(uh*vh*n_Λ)⊙jump(vh*vh*n_Λ))*dΛ
+j(uh,vh) = ∫(mean(∇(vh)*uh)⊙jump((∇(vh)⊙∇(uh))*n_Λ))*dΛ
+
+function _change_input(f,θ,uh)
+  dir = similar(uh.dirichlet_values,eltype(θ))
+  U = uh.fe_space
+  uh = FEFunction(U,θ,dir)
+  T = eltype(θ)
```
amartinhuertas, line 187: Can we add here a comment on why we needed to use the strongly-typed lower-level interfaces of `SparseMatrixAssembler`?

amartinhuertas (changes requested):
Hi @kishore-nori ! thanks for the PR! Only minor comments.

amartinhuertas:
Edit the NEWS.md file, please

src/FESpaces/FEAutodiff.jl
kishore-nori: Done!

src/FESpaces/FEAutodiff.jl
kishore-nori: Done!

kishore-nori:
Hi Alberto, thank you very much for the help, inputs and discussion for the PR, and for reviewing. I fixed the suggested changes and added further details and pushed them.

amartinhuertas:
Hi @kishore-nori ... I am afraid that we are increasing too much testing times in Github Actions with this PR. 

Looking at the logs, we have increased total time from 46 min to 1h 5min approx (by 20 mins approx). 

See attached picture below to identify where the hot spot is. 

Can we try to do a better job here and curate the current set of tests such that we significantly reduce this testing times while still testing the same amount of code base within Gridap kernel? Perhaps we can take a look at code coverage reports (https://app.codecov.io/gh/gridap/Gridap.jl/branch/AD-for-SkeletonIntegration/) to understand which parts of the codes we are going through with each of the tests. 

Screenshot from 2022-06-22 11-43-55

kishore-nori:
Hi Alberto, thank you for the logs. Yeah I realised that the testing times have increased due to the new `jacobian` for `Skeleton` functions in `FEAutodiffTests.jl`. The thing is I put in 5 functionals which reflect most of the complications which are possible in such cases. These cover usual `mean` and `jump` terms of `uh` and `vh`, terms with combinations of `∇(uh)`, `∇(vh)`, `Δ(u)` and Δ(vh). Additionally there are test where `jump` and `mean` have operations of `CellFields` inside. One way that I can reduce the run-time is by have `2x2` grid or even `2x1` instead of current `3x3` grid. I can change this based on your suggestions. I kept `reffe` order at `2` to test even `laplacian`s.

amartinhuertas:
> One way that I can reduce the run-time is by having 2x2 grid or even 2x1 grid instead of current 3x3 grid. I can change this based on your suggestions.

I am not sure this is going to help much. Note in the logs that 97% of the time is compilation time. So I would say it is more about the amount of specialized code for which you are triggering compilation from your tests, not the size of the test cases.

amartinhuertas (approved):
CI tests now take 45 mins. Well done @kishore-nori !

kishore-nori:
That's great! I think the wrapping of `assemble(jacobian(...))` into a function might have also helped.

Issue #805, opened 2022-06-22
Problem with `nothing` tag in Autodiff.jl for mixed spatial and DOF based derivatives  


kishore-nori:
Currently `Gridap` at `src/Arrays/Autodiff.jl` uses `nothing` based `ForwardDiff.Tag` for constructing the `config`s for `ForwardDiff.gradient` and `ForwardDiff.jacobian`, 

https://github.com/gridap/Gridap.jl/blob/78b27c4830d407f05ff07782a2a89d9b16f51670/src/Arrays/Autodiff.jl#L51-L59

this is not a problem for spatial `gradient/laplacian`s of `FEFunction`s, as `ForwardDiff` machinery is not being used inside. Currently, I am working with `CellFields` where the spatial `gradient/laplacian` requires the usage of `ForwardDiff` and the interaction of the `tag` generated in this `ForwardDiff` usage has interaction problem with `nothing` tag from derivatives with respect DOF with respect to the functionals involving the spatial derivative, related to determining the order of the tags. Following is the MWE with only `ForwardDiff`:

```julia
using ForwardDiff
using LinearAlgebra

f(θ,x) = θ⋅x

function test_func(θ)
  _f(x) = f(θ,x)
  x = [0.5,0.6]
  norm(ForwardDiff.gradient(_f,x))
end

θ = [2.0,0.5]
# following the strategy of `Gridap.Arrays` `Autodiff.jl`
cfg = ForwardDiff.GradientConfig(nothing,θ,ForwardDiff.Chunk{length(θ)}())
cfg isa ForwardDiff.GradientConfig{Nothing}
function cfg_type(cfg::ForwardDiff.GradientConfig{T}) where T
  T
end
cfg_type(cfg)

xdual = cfg.duals
ForwardDiff.seed!(xdual, θ, cfg.seeds)

ydual = test_func(xdual) # this is the place where tag order error occurs 
result = similar(θ, ForwardDiff.valtype(ydual))

result = ForwardDiff.extract_gradient!(cfg_type(cfg),result,ydual)
```
`ydual = test_func(xdual)` call results in the following error, this where the interaction of dual numbers from different `ForwardDiff` calls occurs. This is due to one tag being `Nothing`. 

```julia
ERROR: Cannot determine ordering of Dual tags ForwardDiff.Tag{var"#_f#12"{Vector{ForwardDiff.Dual{Nothing, Float64, 2}}}, Float64} and Nothing
Stacktrace:
[...]

PR #806, opened 2022-06-22
Added a dummy tag for ForwardDiff configs being constructed in `Gridap` `Autodiff.jl` to fix issue #805

kishore-nori:


kishore-nori:
After making the last commit of changes the CI takes 
```
Test Summary:  | Pass  Total
FESpaces (2/2) |  855    855
1699.394610 seconds (472.28 M allocations: 26.718 GiB, 0.88% gc time, 99.82% compilation time)
```
with a dummy tag function it took:
```
Test Summary:  | Pass  Total
FESpaces (2/2) |  855    855
516.246620 seconds (387.76 M allocations: 21.898 GiB, 1.95% gc time, 99.66% compilation time)
```
I think this is caused by the tag checking now (also have long tag's like we are ending up here may be has to do something with this, I am not sure, will have to look into this, haven't got much details from profiler). Or it could be I am doing something horribly bad in the last commit (I don't feel so, as the changes have been very minimal, just adding a function to existing structs to extract the tag name while setting up `ForwardDiff` `config`s)

PR #815, opened 2022-07-21
Fixed jacobian for general test basis where dim not same as trial for Skeleton integration terms 

kishore-nori:
* In case of skeleton integration jacobian, earlier code assumed that the cell wise dim of trial and test basis is the same 
* now the code has been rewritten to generalise for case when the above may not be the same 
* relevant tests have passed


amartinhuertas (changes requested):
Hi @kishore-nori ! Thanks for your work. Can you please add an entry to the NEWS.md file?

kishore-nori:
Hi Alberto, updated the NEWS.md file!

PR #834, opened 2022-09-26
added `lastindex` for MultiValue and tests

kishore-nori:
Hi @amartinhuertas and @fverdugo, I have added the `lastindex` method for `MultiValue` for using `[end]` and to resolve ambiguity with `length` upon discussion with @amartinhuertas, as currently `lastindex` falls back to `Number`, returning 1, which is not consistent with `length` for the same argument, the above is a small fix for the same using `length` under the hood.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `lastindex` for `MultiValue`s for consistent usage of `[end]` as per `length`. Since PR [#834](https://github.com/gridap/Gridap.jl/pull/834)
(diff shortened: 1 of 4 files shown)


amartinhuertas:
Hi @kishore-nori! Thanks for the PR. Do not forget to add an entry to the NEWS.md file. You will have to add an unreleased section, as in here https://github.com/gridap/Gridap.jl/blob/f7be57ee6ef4169932b211adc440210bf65864d7/NEWS.md#unreleased

amartinhuertas:
May you please add another entry to the NEWS.md file, section "added", that tells that we added BDM RefFEs in PR https://github.com/gridap/Gridap.jl/pull/823 ? We forgot to do this in a previous PR. Thanks!

Issue #357, opened 2020-08-12
Add DiractDistribution for point "things"


santiagobadia:
A point load can be mathematically represented as a distribution, e.g., a Dirac delta. It breaks the `Gridap` flow, since one cannot use Gauss quadratures and numerical integration (what we usually do in FEM) to compute the integral of f*v in that case.

I don't want to consider hacks, e.g., touch the vector entry in a particular node in which you want to put the force  (assuming the force is on a node).

Instead, e.g., I would consider a constructor for `f = DiracDelta(x_0)` that returns a `Distribution <: Field` that we know cannot be used with integration BUT instead, `integrate(f*v,trian,quad)` returns v(x).

Omega-xyZac:
As an aside to this, how would you do the 'hack' in this case? I would implement the above but I don't think I have enough experience in Gridap just yet.

PR #837, opened 2022-10-12
Adding support to `DiracDelta` for generic points in the domain 

kishore-nori:
Hi @santiagobadia and @amartinhuertas, I added methods to `DiracDelta` to support generic points in the domain, which are not grid entities with tags, would be glad to know your suggestions for improvement and changes. This PR is aimed at resolving the issue [#357]. Given that `CellField`s can now be evaluated at a generic point in the domain, the method make use of this to calculate the integral of the composition of `DiracDelta` with a `CellField`, and also works for generic functions compatible to evaluate at a `Gridap` `Point`. 

So as to conform with `DiracDelta` struct, a `DiscreteModel` is constructed with given set of `Point`s to generate `Triangulation` and `Measure`. If we are not required to conform with the existing framework, we could just store the `Point`s as a struct and evaluate the `CellField` point-wise directly, but the above doesn't create much overhead. And probably the later is not generic enough to be extendable to line and surface forcing terms. 

The usage is same as earlier but the constructor is as follows:

```julia
model = CartesianDiscreteModel((0,1,0,1),(3,3))
p = Point(0.1,0.2)
pvec = [p, 1.5*p]
δ = DiracDelta(model,p)
δ_comb = DiracDelta(model,pvec)
```

[WIP] TO DO: support for line and surface loads! (like presented here https://github.com/gridap/Gridap.jl/issues/408#issuecomment-734929575). I think the line load case is extendable in a straight forward manner using `CartesianDiscreteModel` (but may be needs to be done separately for each line - defined by a pair of points..)

### DIFF
--- a/src/CellData/DiracDeltas.jl
+++ b/src/CellData/DiracDeltas.jl
@@ -99,8 +99,10 @@ function DiracDelta(model::DiscreteModel{D}, p::Point{D,T}) where {D,T}
   cache = _point_to_cell_cache(KDTreeSearch(),trian)
   cell = _point_to_cell!(cache, p)
   trianv = TriangulationView(trian,[cell])
-  pquad = GenericQuadrature([p],[one(T)])
-  pmeas = Measure(CellQuadrature([pquad],[[p]],[[one(T)]],trianv,PhysicalDomain(),PhysicalDomain()))
+  point = [p]
+  weight = [one(T)]
+  pquad = GenericQuadrature(point,weight)
+  pmeas = Measure(CellQuadrature([pquad],[pquad.coordinates],[pquad.weights],trianv,PhysicalDomain(),PhysicalDomain()))
   GenericDiracDelta{0,D,NotGridEntity}(trianv,pmeas)
 end
 
@@ -111,7 +113,7 @@ function DiracDelta(model::DiscreteModel{D}, pvec::Vector{Point{D,T}}) where {D,
   cell_points = collect(values(cell_to_pindices))
   points = map(i->pvec[cell_points[i]], 1:length(cell_ids))
   weights_x_cell = collect.(Fill.(one(T),length.(cell_points)))
-  pquad = map(i -> GenericQuadrature(pvec[cell_points[i]],weights_x_cell[i]), 1:length(cell_ids))
+  pquad = map(i -> GenericQuadrature(points[i],weights_x_cell[i]), 1:length(cell_ids))
   trianv = Triangulation(trian,cell_ids)
   pmeas = Measure(CellQuadrature(pquad,points,weights_x_cell,trianv,PhysicalDomain(),PhysicalDomain()))
   GenericDiracDelta{0,D,NotGridEntity}(trianv,pmeas)
--- a/src/CellData/DiracDeltas.jl
+++ b/src/CellData/DiracDeltas.jl
@@ -88,11 +88,9 @@ end
 # For handling DiracDelta at a generic Point in the domain #
 
 function DiracDelta(x::Point{D,T}, model::DiscreteModel{D}) where {D,T}
-  # check if the point is inside an active cell, as it wouldn't be caught for
-  # user-defined functions (i.e. which are not CellFields)
   trian = Triangulation(model)
   cache1 = _point_to_cell_cache(KDTreeSearch(),trian)
-  cell = _point_to_cell!(cache1, x) # throws error if Point not in domain
+  cell = _point_to_cell!(cache1, x)
   point_grid = UnstructuredGrid([x])
   point_model = UnstructuredDiscreteModel(point_grid)
   point_trian = Triangulation(point_model)
@@ -101,11 +99,9 @@ function DiracDelta(x::Point{D,T}, model::DiscreteModel{D}) where {D,T}
 end
 
 function DiracDelta(v::Vector{Point{D,T}},model::DiscreteModel{D}) where {D,T}
-  # check if the points are inside an active cell
   trian = Triangulation(model)
   cache1 = _point_to_cell_cache(KDTreeSearch(),trian)
   cell = map(x -> _point_to_cell!(cache1, x), v)
-  # throws error if any Point not in domain
   point_grid = UnstructuredGrid(v)
   point_model = UnstructuredDiscreteModel(point_grid)
   point_trian = Triangulation(point_model)
--- a/src/CellData/DiracDeltas.jl
+++ b/src/CellData/DiracDeltas.jl
@@ -1,11 +1,13 @@
 abstract type DiracDeltaSupportType <: GridapType end
-struct IsGridEntity <: DiracDeltaSupportType end
-struct NotGridEntity <: DiracDeltaSupportType end
+abstract type IsGridEntity <: DiracDeltaSupportType end
+abstract type NotGridEntity <: DiracDeltaSupportType end
 
-struct DiracDelta{D,S<:DiracDeltaSupportType} <: GridapType
-  Γ::Triangulation{D}
+struct GenericDiracDelta{D,Dt,S<:DiracDeltaSupportType} <: GridapType
+  Γ::Triangulation{Dt}
   dΓ::Measure
 end
+
+const DiracDelta{D} = GenericDiracDelta{D,D,IsGridEntity}
 
 function DiracDelta{D}(
   model::DiscreteModel,
@@ -27,7 +29,7 @@ function DiracDelta{D}(
   trian = BodyFittedTriangulation(model,face_grid,face_to_bgface)
   Γ = BoundaryTriangulation(trian,glue)
   dΓ = Measure(Γ,degree)
-  DiracDelta{D,IsGridEntity}(Γ,dΓ)
+  DiracDelta{D}(Γ,dΓ)
 end
 
 function DiracDelta{D}(
@@ -68,44 +70,64 @@ function DiracDelta{0}(model::DiscreteModel;tags)
   DiracDelta{0}(model,degree;tags=tags)
 end
 
-function (d::DiracDelta)(f)
+function (d::GenericDiracDelta)(f)
   evaluate(d,f)
 end
 
-function evaluate!(cache,d::DiracDelta,f)
+function evaluate!(cache,d::GenericDiracDelta,f)
   ∫(f)*d.dΓ
 end
 
-function get_triangulation(d::DiracDelta)
+function get_triangulation(d::GenericDiracDelta)
   d.Γ
 end
 
 # For handling DiracDelta at a generic Point in the domain #
 
-function DiracDelta(x::Point{D,T}, model::DiscreteModel{D}) where {D,T}
-  trian = Triangulation(model)
-  cache1 = _point_to_cell_cache(KDTreeSearch(),trian)
-  cell = _point_to_cell!(cache1, x)
-  point_grid = UnstructuredGrid([x])
-  point_model = UnstructuredDiscreteModel(point_grid)
-  point_trian = Triangulation(point_model)
-  dx = Measure(point_trian,1)
-  DiracDelta{0,NotGridEntity}(point_trian,dx)
+function _cell_to_pindex(pvec::Vector{<:Point},trian::Triangulation)
+  cache = _point_to_cell_cache(KDTreeSearch(),trian)
+  cell_to_pindex = Dict{Int, Vector{Int32}}()
+  for i in 1:length(pvec)
+    cell = _point_to_cell!(cache, pvec[i])
+    push!(get!(() -> valtype(cell_to_pindex)[], cell_to_pindex, cell), i)
+  end
+  cell_to_pindex
 end
 
-function DiracDelta(v::Vector{Point{D,T}},model::DiscreteModel{D}) where {D,T}
+function DiracDelta(p::Point{D,T}, model::DiscreteModel{D}) where {D,T}
   trian = Triangulation(model)
-  cache1 = _point_to_cell_cache(KDTreeSearch(),trian)
-  cell = map(x -> _point_to_cell!(cache1, x), v)
-  point_grid = UnstructuredGrid(v)
-  point_model = UnstructuredDiscreteModel(point_grid)
-  point_trian = Triangulation(point_model)
-  dx = Measure(point_trian,1)
-  DiracDelta{0,NotGridEntity}(point_trian,dx)
+  cache = _point_to_cell_cache(KDTreeSearch(),trian)
+  cell = _point_to_cell!(cache, p)
+  trianv = TriangulationView(trian,[cell])
+  pquad = GenericQuadrature([p],[one(T)])
+  pmeas = Measure(CellQuadrature([pquad],[[p]],[[one(T)]],trianv,PhysicalDomain(),PhysicalDomain()))
+  GenericDiracDelta{0,D,NotGridEntity}(trianv,pmeas)
 end
 
-function evaluate!(cache,d::DiracDelta{0,NotGridEntity},f::CellField)
-  d_f = lazy_map(f,d.Γ.grid.node_coordinates)
-  dc = DomainContribution()
-  add_contribution!(dc, d.Γ, d_f)
+function DiracDelta(pvec::Vector{Point{D,T}}, model::DiscreteModel{D}) where {D,T}
+  trian = Triangulation(model)
+  cell_to_pindices = _cell_to_pindex(pvec,trian)
+  cell_ids = collect(keys(cell_to_pindices))
+  cell_points = collect(values(cell_to_pindices))
+  points = map(i->pvec[cell_points[i]], 1:length(cell_ids))
+  weights_x_cell = collect.(Fill.(one(T),length.(cell_points)))
+  pquad = map(i -> GenericQuadrature(pvec[cell_points[i]],weights_x_cell[i]), 1:length(cell_ids))
+  trianv = Triangulation(trian,cell_ids)
+  pmeas = Measure(CellQuadrature(pquad,points,weights_x_cell,trianv,PhysicalDomain(),PhysicalDomain()))
+  GenericDiracDelta{0,D,NotGridEntity}(trianv,pmeas)
 end
+
+# TO DO: point_to_cell search cache needs to reused somehow
+# for Finite Element Function build using a fixed FE basis
+# since the basis remain fixed for each cell unless p or h
+# adapted, we can pre-compute them for given DiracDelta and
+# store them in cache. But for this we have to diverge from
+# the current Struct and add more attributes - like cache
+
+# GenericQuadrature doesn't taken FillArrays only Vectors
+# would be good to add this, as it would be useful here
+# The question is if it creates a problem down somewhere?
+# ofcourse excluding the tests which taken into account the types
+# like test_quadrature functions
+# if type independent problems exist for this generalization then
+# this is a more important problem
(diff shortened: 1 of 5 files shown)


amartinhuertas:
HI @kishore-nori ! Thanks for your work. Can you please add an entry to the `NEWS.md` file pointing to this work ?

test/CellDataTests/DiracDeltasTests.jl
```diff
@@ -37,6 +37,22 @@ degree = 2
 δ = DiracDelta{0}(model,tags=[2,4])
 @test sum(δ(v)) ≈ 5
 
+# Tests for DiracDelta at a generic Point in the domain #
+
+p = VectorValue(0.2,0.3)
```
amartinhuertas, line 42: why do you use `VectorValue` and not `Point`?

test/CellDataTests/DiracDeltasTests.jl
```diff
@@ -37,6 +37,22 @@ degree = 2
...
+δ = DiracDelta(p,model)
+@test sum(δ(v)) == v(p)
```
amartinhuertas, line 44: Replace `==` by `\sim` ?

test/CellDataTests/DiracDeltasTests.jl
```diff
@@ -37,6 +37,22 @@ degree = 2
...
+
+pvec = [p,π*p]
+δ = DiracDelta(pvec,model)
+@test sum(δ(v)) == sum(v(pvec))
```
amartinhuertas, line 48: see https://github.com/gridap/Gridap.jl/pull/837/files#r994035461

test/TensorValuesTests/IndexingTests.jl
```diff
@@ -14,6 +14,7 @@ v = VectorValue{4}(a)
 @test size(v) == (4,)
 @test length(v) == 4
 @test lastindex(v) == length(v)
+@test v[end] == a[end]
```
amartinhuertas, line 17: where do these tests come from? shouldnt be these already in `master` ? (i.e., is your branch up to date with `master`?) they seem to be unrelated to this PR.

test/CellDataTests/DiracDeltasTests.jl
kishore-nori: `Point` is not available for usage here, is has been defined in `Gridap.Fields` so just used `VectorValue` (as it seemed to me that the tests are also following the module hierarchy).

test/CellDataTests/DiracDeltasTests.jl
```diff
@@ -37,6 +37,22 @@ degree = 2
...
+
+# below passes but cannot use FESpace and FEFunction here
```
amartinhuertas, line 50: Not sure why this commented code is here. Should it be here or we can just remove it?

test/TensorValuesTests/IndexingTests.jl
kishore-nori: No, I forgot adding these tests, sorry took the opportunity to add it now (instead of again creating a PR), sorry for the confusion, it is not under PR.

src/CellData/DiracDeltas.jl
```diff
@@ -76,3 +79,33 @@ end
...
+  cache1 = _point_to_cell_cache(KDTreeSearch(),trian)
+  cell = map(x -> _point_to_cell!(cache1, x), v)
+  point_grid = UnstructuredGrid(v)
+  point_model = UnstructuredDiscreteModel(point_grid)
+  point_trian = Triangulation(point_model)
+  dx = Measure(point_trian,1)
+  DiracDelta{0,NotGridEntity}(point_trian,dx)
+end
+
+function evaluate!(cache,d::DiracDelta{0,NotGridEntity},f::CellField)
```
amartinhuertas, line 107: why do we need this specialized version of `evaluate!`. Doesn't the generic one work straight away?

test/CellDataTests/DiracDeltasTests.jl
```diff
@@ -37,6 +37,22 @@ degree = 2
...
+# reffe = ReferenceFE(lagrangian,Float64,3)
+# V = FESpace(model,reffe,conformity=:L2)
+# uh = FEFunction(V,rand(num_free_dofs(V)))
+# @test sum(d(uh)) == uh(p)
+
 #using Gridap
```
amartinhuertas, line 56: should we also test DiracDeltas with `FEFunction` and `FEBasis` ?

test/CellDataTests/DiracDeltasTests.jl
kishore-nori: This is a good test but as I mentioned in the comment, I cant use `FESpace` and `FEFunction` here, as they are down the hierarchy. If it is ok I can use `using Gridap` for the test

amartinhuertas (changes requested):
Hi @kishore-nori ! I have finished a first round of review. Please address or rebut my comments/remarks/etc..

test/CellDataTests/DiracDeltasTests.jl
kishore-nori: Yes, liked I replied to one of the above comments, we need to include the corresponding modules to use the `FEFunction` and `FEBasis`, I would be happy to include these tests..

src/CellData/DiracDeltas.jl
kishore-nori: No.. it results in quadrature and `FEFunction/CellField` not having the same/compatible background triangulation error

test/CellDataTests/DiracDeltasTests.jl
kishore-nori: Yes I can do that, should be safer..

test/CellDataTests/DiracDeltasTests.jl
kishore-nori: (replied above)

test/CellDataTests/DiracDeltasTests.jl
amartinhuertas: >  I cant use FESpace and FEFunction here, as they are down the hierarchy. 

I dont understand this comment. Can't you just add `using Gridap.FESpaces` to the module in that test file?

kishore-nori: Yes I can.., my question was does `Gridap` `tests` also follow the module usage hierarchy as the `src`? If not I can do it.

test/CellDataTests/DiracDeltasTests.jl
amartinhuertas: I dont see any problem in `using Gridap.Fields` at the test. (perhaps I am missing something)

kishore-nori:
Thank you very much for the comments Alberto, I ll make changes accordingly following our discussion in comments here.

test/CellDataTests/DiracDeltasTests.jl
amartinhuertas: I dont think we have to constrain ourselves to the same module usage hierarchy. In Gridap this has to be respected because otherwise we may end with dependcy loops (cycles) among modules.

You can think as the test directory as an external program that is using Gridap tools and features. Thus, I do not see why we are forced to respect that constraint.

src/CellData/DiracDeltas.jl
amartinhuertas: Ok. I now understand. The integrand is defined on a domain which is not (currently) compatible with the domain of the measure. But this may mean that there is a compatibility rule among triangulations that is somehow missing, because why these domains should not be compatible?

amartinhuertas: This is just a though ... it you have the impression that this overly complicates thing, we may stick into the current workaround.

test/TensorValuesTests/IndexingTests.jl
amartinhuertas: ok, no worries ... we may merge them here, but the cleaner way to go is to do it in a separate PR

test/CellDataTests/DiracDeltasTests.jl
amartinhuertas: The other thing that we have to think is whether `DiracDeltas.jl` should still be within `CellData` now given that we have introduced additional dependencies.

test/CellDataTests/DiracDeltasTests.jl
kishore-nori: No problem will use it now, I was in a misconception that even the tests follow the same module include hierarchy as the Gridap source code does.

test/TensorValuesTests/IndexingTests.jl
kishore-nori: Yes I am sorry, will be careful next time and separate it into different PR. Probably would be too messy here now, to revert back?

src/CellData/DiracDeltas.jl
kishore-nori: Initially I tried to make it work with the same triangulations but it wasn't generalising properly due to `CellQuadrature` constructing extra-cells of `#undef` values in `DomainContribution`. 

The problem occurs at @check is_change_possible(trian_f,trian_x) msg in the integrate method (https://github.com/gridap/Gridap.jl/blob/2a4921b473fe113b9d4b6168b8728a7a708f7bc6/src/CellData/CellQuadratures.jl#L142)

kishore-nori: > But this may mean that there is a compatibility rule among triangulations that is somehow missing, because why these domains should not be compatible?
Yes I was wondering about this as well.. will have to dig deeper

test/CellDataTests/DiracDeltasTests.jl
kishore-nori: > The other thing that we have to think is whether DiracDeltas.jl should still be within CellData now given that we have introduced additional dependencies.
No.. I haven't introduced any dependency in DiracDeltas.jl, used only the existing dependencies

amartinhuertas: > No.. I haven't introduced any new dependency in DiracDeltas.jl, used only the existing dependencies

ok, I think you are right. Having a (new) reference to `CellField` (in the new `evaluate` function) does not imply a new module dependency, correct?

test/TensorValuesTests/IndexingTests.jl
amartinhuertas: Better not to revert back, I agree. 

For completeness, perhaps we can add in the NEWS.md a comment that we added some missing tests missing from another PR.

test/CellDataTests/DiracDeltasTests.jl
kishore-nori: Yes.. since `CellField` belongs to `CellData` module itself, it is not a new or external dependency.

test/TensorValuesTests/IndexingTests.jl
kishore-nori: Definitely, will mention it in the `NEWS.md`

kishore-nori:
Hi @amartinhuertas and @santiagobadia, I have rewritten the implementation (realised I could do it in a good way, after our discussion to approach it as quadrature). This now works for evaluations and also solving the PDE with `DiracDelta` at a single or even a set of `Points`. I added tests to compare with both the exact solution and the earlier tag based implementation of `DiracDelta`. I also incorporated the earlier review comments from our discussions. Would be happy to know further comments for improvement.

src/CellData/DiracDeltas.jl
```diff
@@ -65,14 +70,63 @@ function DiracDelta{0}(model::DiscreteModel;tags)
...
+  cell_points = collect(values(cell_to_pindices))
+  points = map(i->pvec[cell_points[i]], 1:length(cell_ids))
+  weights_x_cell = collect.(Fill.(one(T),length.(cell_points)))
+  pquad = map(i -> GenericQuadrature(pvec[cell_points[i]],weights_x_cell[i]), 1:length(cell_ids))
+  trianv = Triangulation(trian,cell_ids)
+  pmeas = Measure(CellQuadrature(pquad,points,weights_x_cell,trianv,PhysicalDomain(),PhysicalDomain()))
+  GenericDiracDelta{0,D,NotGridEntity}(trianv,pmeas)
+end
+
+function evaluate!(cache,d::GenericDiracDelta{0,Dt,NotGridEntity},f::Function) where Dt
```
amartinhuertas, line 120: Which is the ultimate reason behind having to define this specialised method? 

Why cant we leverage the generic implementation of `evaluate!` whenever the domain of `DiracDelta` is a function?

src/CellData/DiracDeltas.jl
```diff
@@ -65,14 +70,63 @@ function DiracDelta{0}(model::DiscreteModel;tags)
...
+  cell_to_pindex
+end
+
+function DiracDelta(model::DiscreteModel{D}, p::Point{D,T}) where {D,T}
+  trian = Triangulation(model)
+  cache = _point_to_cell_cache(KDTreeSearch(),trian)
+  cell = _point_to_cell!(cache, p)
+  trianv = TriangulationView(trian,[cell])
+  pquad = GenericQuadrature([p],[one(T)])
+  pmeas = Measure(CellQuadrature([pquad],[[p]],[[one(T)]],trianv,PhysicalDomain(),PhysicalDomain()))
```
amartinhuertas, line 103: to be on the safe side w.r.t. memory allocations, better not to create `[p]` and `[one(T)]` twice.

why do we have to pass `[[p]]` to `CellQuadrature` constructor? Couldn't it get this info from `pquad` ? (perhaps a missing constructor in Gridap?)

amartinhuertas (changes requested):
Hi @kishore-nori ! nice work! I finished with my review. Please answer/rebut my comments.

src/CellData/DiracDeltas.jl
kishore-nori: I was in a misconception that existing `evaluate!` method goes through the search process of Point to Cell (which ofcourse works perfectly fine as well), which is only necessary for `CellFields`, Functions can be evaluated directly without the search. But it seems it doesn't go through that route once we have the `CellPoint` available, which is very nice. So I ll remove it..

src/CellData/DiracDeltas.jl
kishore-nori: Yes true, I was thinking the same, may be a missing constructor, specifically for `data_domain_style` being `PhysicalDomain()`, the problem is we can't use the `compress_cell_data` for `AbstractArrays` only for `Fill` and `CompressedArray`, which is the usual quadrature case. Yes I ll remove the repetition and reuse.

src/CellData/DiracDeltas.jl
kishore-nori: But one advantage of the new constructor is it is ~4 times faster and aloocates ~10 times less memory.

amartinhuertas: > But it seems it doesn't go through that route once we have the CellPoint available, which is very nice.

Whenever you have a `CellPoint`, Gridap assumes that the points in each cell are within the cell (either reference cell or physical cells, depending on the value of the corresponding trait). 

If the points are not within the cell, then I guess you have an undefined behaviour. But we are satisfying this precondition, as the cells have been judiciously selected in the TriangulationView such that the hold the corresponding points.

amartinhuertas: > But one advantage of the new constructor is it is ~4 times faster and aloocates ~10 times less memory.

Do these ratios keep asympotatically fixed as you refine the mesh?

kishore-nori: > Whenever you have a CellPoint, Gridap assumes that the points in each cell are within the cell (either reference cell or physical cells, depending on the value of the corresponding trait).

>If the points are not within the cell, then I guess you have an undefined behaviour. But we are satisfying this precondition, as the cells have been judiciously selected in the TriangulationView such that they hold the corresponding points.

Yes true.. since we are checking it, it should be safe, it is just a matter of efficiency

kishore-nori: > Do these ratios keep asympotatically fixed as you refine the mesh?
I ll check and update.

kishore-nori: On refinement the the new version runs in the same time but the earlier one becomes slower and more memory intensive.

src/CellData/DiracDeltas.jl
amartinhuertas: > Yes true, I was thinking the same, may be a missing constructor, specifically for data_domain_style being PhysicalDomain(), the problem is we can't use the compress_cell_data for AbstractArrays only for Fill and CompressedArray, which is the usual quadrature case. Yes I ll remove the repetition and reuse.

Ok, I see. The constructor is missing, but if we added it, we would not be able to leverage it in our particular context (we do not have Fill or CompressedArrays to represent the quadrature points and weights).

src/CellData/DiracDeltas.jl
```diff
@@ -65,14 +70,65 @@ function DiracDelta{0}(model::DiscreteModel;tags)
...
+    push!(get!(() -> valtype(cell_to_pindex)[], cell_to_pindex, cell), i)
+  end
+  cell_to_pindex
+end
+
+function DiracDelta(model::DiscreteModel{D}, p::Point{D,T}) where {D,T}
+  trian = Triangulation(model)
+  cache = _point_to_cell_cache(KDTreeSearch(),trian)
+  cell = _point_to_cell!(cache, p)
+  trianv = TriangulationView(trian,[cell])
```
amartinhuertas, line 101: use `view` Julia method instead of `TriangulationView`. 

This way we do not need to export `TriangulationView`.

src/Geometry/Geometry.jl
```diff
@@ -102,6 +102,7 @@ export compress_ids
 export UnstructuredGridTopology
 
 export Triangulation
+export TriangulationView
```
amartinhuertas, line 105: see https://github.com/gridap/Gridap.jl/pull/837/files#r996625877

src/CellData/DiracDeltas.jl
```diff
@@ -65,14 +70,65 @@ function DiracDelta{0}(model::DiscreteModel;tags)
...
+
+function DiracDelta(model::DiscreteModel{D}, pvec::Vector{Point{D,T}}) where {D,T}
+  trian = Triangulation(model)
+  cell_to_pindices = _cell_to_pindices(pvec,trian)
+  cell_ids = collect(keys(cell_to_pindices))
+  cell_points = collect(values(cell_to_pindices))
+  points = map(i->pvec[cell_points[i]], 1:length(cell_ids))
+  weights_x_cell = collect.(Fill.(one(T),length.(cell_points)))
+  pquad = map(i -> GenericQuadrature(points[i],weights_x_cell[i]), 1:length(cell_ids))
+  trianv = TriangulationView(trian,cell_ids)
```
amartinhuertas, line 117: https://github.com/gridap/Gridap.jl/pull/837/files#r996625877

src/CellData/DiracDeltas.jl
```diff
@@ -65,14 +70,65 @@ function DiracDelta{0}(model::DiscreteModel;tags)
...
+  pmeas = Measure(CellQuadrature(pquad,points,weights_x_cell,trianv,PhysicalDomain(),PhysicalDomain()))
+  GenericDiracDelta{0,D,NotGridEntity}(trianv,pmeas)
+end
+
+function evaluate!(cache,d::GenericDiracDelta{0,Dt,NotGridEntity},f::Function) where Dt
```
amartinhuertas, line 122: Remove this method.

src/Geometry/Geometry.jl
kishore-nori: done!

src/CellData/DiracDeltas.jl
kishore-nori: done!

src/CellData/DiracDeltas.jl
kishore-nori: done :)

Issue #789, opened 2022-05-26
Possibility of doing all computations using `Float32` for performance


kishore-nori:
This is some notes from my discussions with @santiagobadia  and @amartinhuertas, regarding the option to use `Float32` everywhere. Currently there are places where `Float64` is hard-coded, especially the quadrature. 

- For example, we can see in `/src/ReferenceFEs/TensorProductQuadratures.jl`, that `T=Float64` is hard-coded into `_tensor_product_legendre(degrees)` function. (`gauss` quadrature from `QuadGK.jl`, which is used here, can handle `Float32`)
- Both `CartesianDiscreteModel{D,T,F}` and `UnstructuredDiscreteModel{Dc,Dp,Tp,B}` carry Floating-point type parameter `T` and `Tp` respectively, which can be propagated into quadrature constructors. But we note that their supertype `DiscreteModel` doesn't have this type parameter
- Even if the model itself is in `Float64`, we can have the option of quadrature to be `Float32`, in cases where we are not using FE basis but only using `Gridap` as means to integrate some external functions.

PR #840, opened 2022-10-19
Adding support for quadrature in `Float32` precision

kishore-nori:
Hi @amartinhuertas and @fverdugo, I have added support to include a keyword argument `T` into `Measure`, `CellQuadrature` and `Quadrature`, to perform numerical integration in the desired floating-point precision, in particular, `Float32` usage. Tests have been added to for checks with `Float32` option for all the effected modules.

This partially resolves the issue https://github.com/gridap/Gridap.jl/issues/789 for the quadrature part. The setup of `FESpace` at places seems hard-coded in `Float64`, and the floating-point type is not being taken from `model` and `reffe` constructed using `Float32` precisions, which I leave for later. 

This doesn't break the high-level API, as the optional `kwarg` `T` defaults to `Float64`. The usage is as follows:
```julia
dΩ = Measure(Ω,2,T=Float32)
quad_strang = Quadrature(TET,2,T=Float32)
```

Would be happy to know to comments and suggestions for changes and improvements.

### DIFF
--- a/src/ReferenceFEs/DuffyQuadratures.jl
+++ b/src/ReferenceFEs/DuffyQuadratures.jl
@@ -25,11 +25,11 @@ function _duffy_quad_data(order::Integer,D::Int;fptype::Type{<:AbstractFloat}=Fl
   dim_to_xs_1d = [quad_1d[x_pos] for quad_1d in dim_to_quad_1d]
   dim_to_ws_1d = [quad_1d[w_pos] for quad_1d in dim_to_quad_1d]
 
-  a = 0.5
+  a = fptype(0.5)
   for d in (D-1):-1:1
     ws_1d = dim_to_ws_1d[d]
     ws_1d[:] *= a
-    a *= 0.5
+    a *= fptype(0.5)
   end
 
   x,w = _tensor_product_duffy(dim_to_xs_1d,dim_to_ws_1d)
@@ -96,8 +96,9 @@ end
 function _tensor_product_duffy!(
   xs,ws,dim_to_xs_1d,dim_to_ws_1d,cis::CartesianIndices{D},m) where D
   k = 1
+  T = eltype(ws)
   for ci in cis
-    w = 1.0
+    w = one(T)
     for d in 1:D
       xs_1d = dim_to_xs_1d[d]
       ws_1d = dim_to_ws_1d[d]
--- a/src/ReferenceFEs/DuffyQuadratures.jl
+++ b/src/ReferenceFEs/DuffyQuadratures.jl
@@ -3,21 +3,21 @@ struct Duffy <: QuadratureName end
 
 const duffy = Duffy()
 
-function Quadrature(p::Polytope,name::Duffy,degree::Integer;fptype::Type{<:AbstractFloat}=Float64)
+function Quadrature(p::Polytope,name::Duffy,degree::Integer;T::Type{<:AbstractFloat}=Float64)
   @assert is_simplex(p)
   D = num_dims(p)
-  x,w = _duffy_quad_data(degree,D,fptype=fptype)
+  x,w = _duffy_quad_data(degree,D,T=T)
   msg = "Simplex quadrature of degree $degree obtained with Duffy transformation and tensor product of 1d Gauss-Jacobi and Gauss-Legendre rules."
   GenericQuadrature(x,w,msg)
 end
 
-function _duffy_quad_data(order::Integer,D::Int;fptype::Type{<:AbstractFloat}=Float64)
+function _duffy_quad_data(order::Integer,D::Int;T::Type{<:AbstractFloat}=Float64)
 
   beta = 0
   dim_to_quad_1d = [
-    _gauss_jacobi_in_0_to_1(order,(D-1)-(d-1),beta;fptype=fptype) for d in 1:(D-1) ]
+    _gauss_jacobi_in_0_to_1(order,(D-1)-(d-1),beta;T=T) for d in 1:(D-1) ]
 
-  quad_1d = _gauss_legendre_in_0_to_1(order;fptype=fptype)
+  quad_1d = _gauss_legendre_in_0_to_1(order;T=T)
   push!(dim_to_quad_1d,quad_1d)
 
   x_pos = 1
@@ -25,11 +25,11 @@ function _duffy_quad_data(order::Integer,D::Int;fptype::Type{<:AbstractFloat}=Fl
   dim_to_xs_1d = [quad_1d[x_pos] for quad_1d in dim_to_quad_1d]
   dim_to_ws_1d = [quad_1d[w_pos] for quad_1d in dim_to_quad_1d]
 
-  a = fptype(0.5)
+  a = T(0.5)
   for d in (D-1):-1:1
     ws_1d = dim_to_ws_1d[d]
     ws_1d[:] *= a
-    a *= fptype(0.5)
+    a *= T(0.5)
   end
 
   x,w = _tensor_product_duffy(dim_to_xs_1d,dim_to_ws_1d)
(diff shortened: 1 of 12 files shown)


amartinhuertas:
@kishore-nori !  Thanks for your work. Can you please add an entry to the NEWS.md file?

kishore-nori:
Thank you for checking Alberto. Done! (sorry forgot `[skip ci]`).

src/ReferenceFEs/TensorProductQuadratures.jl
```diff
@@ -3,24 +3,24 @@ struct TensorProduct <: QuadratureName end
...
+  Quadrature(p,name,degrees,fptype=fptype)
 end
 
 # Low level constructor
 
-function _tensor_product_legendre(degrees)
+function _tensor_product_legendre(degrees;T::Type{<:AbstractFloat}=Float64)
     D = length(degrees)
-    T = Float64
+    # T = Float64
```
amartinhuertas, line 23: Remove that line. Do not comment it out.

kishore-nori: Sure.. will remove it.

kishore-nori: Done!

src/ReferenceFEs/DuffyQuadratures.jl
```diff
@@ -52,22 +52,24 @@ end
...
 end
 
 # Transforms a 1-D quadrature from `[-1,1]` to `[a,b]`, with `a<b`.
-function _map_to(a,b,points,weights)
-  points_ab = 0.5*(b-a)*points .+ 0.5*(a+b)
-  weights_ab = 0.5*(b-a)*weights
+function _map_to(a,b,points,weights;T::Type{<:AbstractFloat}=Float64)
+  points_ab = Vector{T}(undef,length(points))
+  weights_ab = Vector{T}(undef,length(weights))
+  points_ab .= 0.5*(b-a)*points .+ 0.5*(a+b)
```
amartinhuertas, line 71: Here we might be mixing different floating point precision numbers.  

E.g. `0.5` is `Float64` while `points_ab` eltype might be, say, `Float32`. 

Thus, a number of implicit conversions are being performed in that case. 

I wonder which is the performance impact of these conversions and whether it is convenient to be systematic in performing conversions so that we do not have many implicit conversions.

src/ReferenceFEs/StrangQuadratures.jl
```diff
@@ -72,7 +72,7 @@ function _strang_quad_tri(degree)
     et2 = 0.445948490915965
     ez2 = 0.445948490915965
     a = 0.054975870996713638
-    b = 0.1116907969117165    
+    b = 0.1116907969117165
```
amartinhuertas, line 75: see https://github.com/gridap/Gridap.jl/pull/840/files#r999428126. `b` is `Float64`

src/ReferenceFEs/DuffyQuadratures.jl
kishore-nori: You are right! I felt implicit conversion is more efficient than first forming the Vector in `Float64` (as the values coming from `gausslegendre` and `gaussjacobi` are of `Float64` eltype) and then constructing a new vector of  converted `Float32` values. To handle `0.5` I can put `T` around it, if you also think it would be good idea.

src/ReferenceFEs/StrangQuadratures.jl
kishore-nori: Yes, implicit conversion I felt (like commented above) to be more efficient than converting the whole vector of `Float64`. A tedious way would be to wrap `fptype` around all these values, I feel the current way and this way internally do the same thing of converting `Float64` values to `Float32`.

src/ReferenceFEs/DuffyQuadratures.jl
kishore-nori: I tested a bit on such situations and it doesn't seem necessary to convert the RHS to the floating point type of LHS and the implicit conversion is slightly faster. I added places where the usage of stated floating precision for LHS was missing (very minor ones actually). I think it is better to have consistent floating point precision for elements in RHS (doesn't have to be the same as LHS if implicit conversion is possible!)

amartinhuertas: > **it doesn't seem necessary** to convert the RHS to the floating point type of LHS

You mean explicit conversion, correct? I indeed see the need to perform the conversion (no matter whether it is generated by the compiler implicitly or by the programmer explicitly) if the `eltype`s of both sides do not match.

Something we could also do is to force that `a`, `b`, `points` and `weights` are of the same type and perform the conversion (explicitly) before calling the function. This way you do not have to pass optional `T`. What do you think?

amartinhuertas: > (as the values coming from gausslegendre and gaussjacobi are of Float64 eltype)

We would extend `gausslegendre` and `gaussjacobi` with a new method each that takes an extra argument with `fptype`  and internally calls the current `gausslegendre` and `gaussjacobi` and performs the explicit conversion.

src/ReferenceFEs/StrangQuadratures.jl
```diff
@@ -24,7 +24,7 @@ end
 
 const _strang_tri_k2n = Dict(1=>1,2=>3,3=>4,4=>6,5=>7,7=>13,9=>19,11=>28)
 
-function _strang_quad_tri(degree)
+function _strang_quad_tri(degree;T::Type{<:AbstractFloat}=Float64)
```
amartinhuertas, line 27: Why do you use `T` instead of `fptype`? I would try to stick to the same name for consistency reasons. 

On the other hand, I think `T` is a quite standard name for eltypes in Gridap and Julia. Not sure about `fptype`. What do you think?

src/ReferenceFEs/DuffyQuadratures.jl
kishore-nori: > You mean explicit conversion, correct? I indeed see the need to perform the conversion (no matter whether it is generated by the compiler implicitly or by the programmer explicitly) if the eltypes of both sides do not match.

yes true.. I meant explicit conversion. 

> Something we could also do is to force that a, b, points and weights are of the same type and perform the conversion (explicitly) before calling the function. This way you do not have to pass optional T. What do you think?

This could be done but this incurs additional memory to store the new vectors of converted type, anyways I feel since `a`, `b`, `points` and `weights` are all in `Float64`, the RHS happens consistently in `Float64` without needing conversion until the storage into LHS, this helps in not having internal floating point cancellations in RHS calculation due to a lower precision arithmetic.

kishore-nori: > We would extend gausslegendre and gaussjacobi with a new method each that takes an extra argument with fptype and internally calls the current gausslegendre and gaussjacobi and performs the explicit conversion.

Like mentioned above this might lead to degraded prec even at the RHS calculation stage, downgrading to lower precision can be done only at the storage point, anyways the scaling (which is happening in the RHS here) doesn't bring in much computation difference even done in `Float32` as it is being done only once when quadrature is setup, and might lead to lower accuracy due to internal cancellations, I feel.

src/ReferenceFEs/StrangQuadratures.jl
kishore-nori: yes sure I can continue to use `fptype` instead of `T`, in other internal functions I had used `T` because `T` was already being used, for example, `TensorProductQuadratures`.

Yes I thought about it, but the problem is as a keyword argument `T` doesn't convey much information to the user.

amartinhuertas: > Yes I thought about it, but the problem is as a keyword argument T doesn't convey much information to the user.

I don't agree here. To a reasonably experienced Julia programmer, `T` tells a lot. 

`T` is the canonical name for the `eltype` type parameter in `Array{T,N}` and in many other types. 

If instead of `T` we called it, say, `H`, I would agree with you.

kishore-nori: True.., I can make it `T` everywhere if that is better?

src/ReferenceFEs/DuffyQuadratures.jl
amartinhuertas: Ok, I think the argument of degraded precision might be a good one. The strategy would be to perform the computations leading to the final weights and points in `Float64`, and defer the conversion to the very last moment, i.e., when you have to store the result into the corresponding arrays of the desired precision. But isn't this contradictory with the following you did?

> on the other hand gauss from QuadGK.jl has this option and I used it.

kishore-nori: True, rightly pointed out the inconsistency, I aware of this but even then I did this to avoid extra storage for conversion and also because the manipulation after the gauss points and weights are generated is very little and mostly trivial (`1.0` and `0.5` have exact storage in both precision!)

src/ReferenceFEs/StrangQuadratures.jl
kishore-nori: sure will do it now, along with any other changes, thank you for the decision..

kishore-nori: Done!

src/ReferenceFEs/DuffyQuadratures.jl
kishore-nori: Yes Alberto..

amartinhuertas: > Yes Alberto..

ok, perfect.

amartinhuertas (approved):

amartinhuertas:
@fverdugo ... do you want to comment on this PR? Thanks!

PR #849, opened 2022-11-29
Fix collect matrix bug

oriolcg:
Hi @fverdugo, I fixed a bug that I encountered when trying to assemble a rectangular matrix. I added a test that stresses this feature.

### DIFF
--- a/test/FESpacesTests/AssemblersTests.jl
+++ b/test/FESpacesTests/AssemblersTests.jl
@@ -100,6 +100,22 @@ assemble_matrix_and_vector!(a(du,dv),ℓ(dv),A22,b22,U,V)
 @test norm(A22-A2) < tol
 @test norm(b22-b2) < tol
 
+# Rectangular matrix
+a2(u,v) = ∫(u⋅∇(v))*dΩ
+V2 = TestFESpace(
+  model,
+  ReferenceFE(lagrangian,VectorValue{2,Float64},1))
+U2 = TrialFESpace(V2)
+
+du2 = get_trial_fe_basis(U2)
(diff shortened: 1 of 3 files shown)


test/FESpacesTests/AssemblersTests.jl
```diff
@@ -100,6 +100,22 @@ assemble_matrix_and_vector!(a(du,dv),ℓ(dv),A22,b22,U,V)
...
+  model,
+  ReferenceFE(lagrangian,VectorValue{2,Float64},1))
+U2 = TrialFESpace(V2)
+
+du2 = get_trial_fe_basis(U2)
+mat_contribs = a2(du2,dv)
+vec_contribs = ℓ(dv)
+
+assem = SparseMatrixAssembler(U2,V)
+data = collect_cell_matrix(U2,V,mat_contribs)
```
amartinhuertas, line 115: Why `collect_cell_matrix` and not `collect_cell_matrix_and_vector` ? The BUG fix was in the second function, right? (most probably I am missing something).

oriolcg: You're right. Added some additional calls that stress the `collect_cell_matrix_and_vector` function.

amartinhuertas: > You're right. I added some additional calls that stress the collect_cell_matrix_and_vector function.

Thanks for your work. A couple of extra points:
 1. Does the `collect_cell_matrix` function also have the bug? If yes, did you fix it as well? I only see bug-fixes in `collect_cell_matrix_and_vector`.
 2. Please dont forget to update NEWS.md.

PR #851, opened 2022-12-01
Helper for loading gmsh

antoine-levitt:


src/Geometry/DiscreteModels.jl
```diff
@@ -415,12 +415,11 @@ function DiscreteModelFromFile(filename::AbstractString)
 end
 
 function DiscreteModelFromFile(filename::AbstractString,::Any)
-  @notimplemented
+  error("File type unknown (did you forget to load GridapGmsh?)")
```
fverdugo, line 418: Can you check that the extension in `filename` is indeed `.msh` to include the hint about loading GridapGmsh ?

fverdugo (changes requested):
Hi @antoine-levitt,
thanks for this PR!
See the very minor comment below.

src/Geometry/DiscreteModels.jl
antoine-levitt: I adapted the warning to be more generic

PR #856, opened 2022-12-13
Add `ascii` and `append` kwargs to writevtk interface

pmartorell:
Added `ascii = false, append = ! ascii` Bool kwargs to writevtk.

The defaults are set according to `WriteVTK.jl` manual:

https://jipolanco.github.io/WriteVTK.jl/v1.16/grids/syntax/

As `ascii` is not compatible with `append` the default append is set to `! ascii`

### DIFF
--- a/src/Visualization/Vtk.jl
+++ b/src/Visualization/Vtk.jl
@@ -1,20 +1,22 @@
 """
 """
-function writevtk(args...;kwargs...)
+function writevtk(args...;ascii=false,append=!ascii,kwargs...)
   map(visualization_data(args...;kwargs...)) do visdata
     write_vtk_file(
-    visdata.grid,visdata.filebase,celldata=visdata.celldata,nodaldata=visdata.nodaldata)
+      visdata.grid,visdata.filebase;
+      celldata=visdata.celldata,nodaldata=visdata.nodaldata,ascii,append)
   end
 end
 
 """
 """
-function createvtk(args...;kwargs...)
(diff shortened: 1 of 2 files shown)


fverdugo:
Hi @pmartorell 

thanks for this.

Up to know the options we were (implicitly) using were ` ascii=false, append=false`, right? Why to change them to ` ascii=false, append=true`

pmartorell:
No, the default options of `vtk_grid` are `ascii=false, append=true`. 

Check the following, be aware that the example do not show the defaults. The defaults are written in bullet points.

https://jipolanco.github.io/WriteVTK.jl/v1.16/grids/syntax/#Supported-options

`vtk_grid`  creates `DatesetFile` with the stated defaults, see:

https://github.com/jipolanco/WriteVTK.jl/blob/55dd5d97644c2fee761a04ea20f95e815b05ad9c/src/WriteVTK.jl#L63

In summary the behaviour (of the proposed implementation) is the following:
*  `(;)` -> `ascii=false,append=true` : `vtk_grid` default, writes appended binary
* `(;append=false)`  -> `ascii=false,append=false` : writes not appended binary
* ` (;ascii=true)` -> `ascii=true,append=false` : writes in ASCII
* `(;ascii=true,append=true) ` -> `ascii=true,append=true` : `ascii` is ignored, writes appended binary

JordiManyer:
All kwargs were added in PR #1016 , which includes this PR.

PR #858, opened 2022-12-22
Using broadcasting in ThetaMethod updates

amartinhuertas:


oriolcg:
Hi @amartinhuertas, would it be an option to use the `@.` macro?

Also, this change also applies to other time integration schemes, right? I can do the changes there.

amartinhuertas:
> would it be an option to use the @. macro?

Definitely yes. I always forget about the macro. Thanks for pointing this out!

> Also, this change also applies to other time integration schemes, right? I can do the changes there.

Absolutely yes. I guess there will be other ode solvers where this may have been overlooked (resulting in frequent temporary memory allocation of system vectors). If you can take a look at that it would be great (I am not that familiar with the ODEs module of Gridap).

PR #869, opened 2023-02-01
Fix SubArray issue for MultiField AutoDiff

JanModderman:
Fix to an issue arising when using AutoDiff with MultiFieldFE spaces. The view() function would return a SubArray, causing an error referred to as map-type unstable.

### DIFF
--- a/src/MultiField/MultiFieldFEAutodiff.jl
+++ b/src/MultiField/MultiFieldFEAutodiff.jl
@@ -87,7 +87,7 @@ function FESpaces._change_argument(
     cell_dofs_field_offsets=_get_cell_dofs_field_offsets(uh)
     for i in 1:nfields
       view_range=cell_dofs_field_offsets[i]:cell_dofs_field_offsets[i+1]-1
-      cell_values_field = lazy_map(a->view(a,view_range),cell_u)
+      cell_values_field = lazy_map(a->a[view_range],cell_u)
       cf = CellField(U.spaces[i],cell_values_field)
       push!(single_fields,cf)
     end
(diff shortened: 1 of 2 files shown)


JanModderman:
Added the reproducer file to the tests directory, please let me know if I need to do anything else as this is my first time doing a pull request.

fverdugo:
I just merged, but I have realized that your reproducer is not being executed. Can you please include it here:

https://github.com/gridap/Gridap.jl/blob/7f14e4e1a1ece7b2ce8fc70d0bd586065e4ef1d8/test/GridapTests/runtests.jl#L48

Thanks!

amartinhuertas:
> Fix to an issue arising when using AutoDiff with MultiFieldFE spaces. The view() function would return a SubArray, causing an error referred to as map-type unstable.

Please note that I used view() instead of array indexing to avoid dynamic memory allocation. 

Which is the error that you get when using view() ?

fverdugo:
Yes indeed. With this now the code works. the next step would be to make it work more efficiently.

The problem is that at some point we call b = similar(a,T,n) with a::SubArray which does not return a SubArray instance, but we need typeof(a) == tupeof(b).

@JanModderman can you point in which line was the problematic call to similar? I remember it was in some function called pos_neg_values or something like this.

JanModderman:
As for running the test @fverdugo, I think I need to load the GridapEmbedded package to reproduce the error, but I can't do this within the GridapTests. How should I proceed here?

Currently at a seminar so I'll be available to work on this afternoon/tomorrow.

fverdugo:
I think you don't really need GridapEmbedded to reproduce the error. 

Simply define your fespaces in a triangulation \Omega_act = Interior(model,mask) that is a sub-set of the \Omega = Interior(model). And integrate the residual in \Omega. For mask you can use for instance mask=map(isodd,1:num_cells(model))

PR #870, opened 2023-02-02
Added reproducer and tests for Issue 869

janmodderman:
Hi @fverdugo, this should conclude Issue 869.


fverdugo:
Thanks. It looks good!

Can you please confirm that your reproducer fails with the last registered version?

JanModderman:
I still get map-type instable when using Gridap@0.17.16

JanModderman:
https://pastebin.com/C8Dyz6yL the error for completeness

PR #872, opened 2023-02-06
Fix of ConstantFESpace + tests + export

janmodderman:
Hi,

I added the function ConstantFESpace to the export, included the test file to the run_test file and fixed some minor things in the ConstantFESpace file. Also, I wasn't sure what the package GridapBiotElasticity is in the test file, so I just commented it out, it runs regardless.

Best regards,

Jan


amartinhuertas (approved):
Hi @janmodderman ! Thanks for the PR! GridapBiotElasticity is a package where this FEspace was originally developed, and thus should not be in Gridap. I have removed the comment.

Issue #877, opened 2023-02-21
Select quadrature data domain


ericneiva:
Hi, @amartinhuertas,

Some time ago, you added in a6cc908d a trait template parameter to `CellQuadrature`. It lets the user select whether to _integrate_ in reference or physical domain. 

Conversely, the default domain for the quad _data_ is reference and it cannot be changed at the user level :point_down:

https://github.com/gridap/Gridap.jl/blob/a6cc908da289ce453413b069d2bc4151fcc628b6/src/CellData/CellQuadratures.jl#L49-L57

The method above is the only one that calls the constructor (see line 56). As you can see it sets the data domain style to `ReferenceDomain()`.

I am interfacing Gridap to a library that computes quad points and weights directly in the physical domain. I can make it work, but I need to replicate the above method to set the data domain to physical.

Have you thought about letting the user have power over both data and integration domain styles? I thought I could change the API of `CellQuadrature` such that both domain styles are keyword arguments. What do you think?

amartinhuertas:
Hi @ericneiva,

thanks for your issue, and sorry for the very late reply (the last months have been quite hectic for me, but anyway ...) 

Let me answer to your comments one by one. 

> The method above is the only one that calls the constructor (see line 56). As you can see it sets the data domain style to ReferenceDomain().

Let me note that apart from https://github.com/gridap/Gridap.jl/commit/a6cc908da289ce453413b069d2bc4151fcc628b6, there is another related commit, namely, https://github.com/gridap/Gridap.jl/commit/5167a37e2458140df80ac9434722a7083e3f1388. In this later commit, I recovered the signature of an older outer constructor (for backward compatibility). This outer constructor also calls the inner constructor, and lets you specify `DataDomainStyle`. In any case, this outer constructor does not provide extra expressivity **as it is possible to call the inner constructor passing all the fields of the structure.** (i.e., the default inner constructor).

> I am interfacing Gridap to a library that computes quad points and weights directly in the physical domain. I can make it work, but I need to replicate the above method to set the data domain to physical.

If am not missing anything, I don't see why there is **a need**.  You can still leverage the default inner constructor of `CellQuadrature` from `Measure`, e.g.:

`Measure(cell_quad, cell_point, cell_weight, trian, PhysicalDomain(), PhysicalDomain())`

> Have you thought about letting the user have power over both data and integration domain styles? I thought I could change the API of CellQuadrature such that both domain styles are keyword arguments. What do you think?

[...]

PR #885, opened 2023-03-13
Customisable measure data domain style

ericneiva:
Solves #877.

After discussion with @santiagobadia, I have changed the `CellQuadrature` constructors such that now both data and integration domain styles are keyword arguments that the user can customise.

Note that this PR introduces a minor change in the API.


amartinhuertas:
Hi @ericneiva ! Thanks for the PR. I agree the API is much cleaner now.

My only concern is that if we accept the PR as it is, then we should also do major release, correct?

Can we may be add the old signatures and add corresponding deprecated warnings so that we avoid a major release? 

Also note that, since this PR (my fault sorry), we also added another kind of measure, called `CompositeMeasure`, which is leveraged whenever you have integration on adapted meshes. Not sure if this has to be taken into account when updating this branch from `master` ...

amartinhuertas:
> Also note that, since this PR (my fault sorry), we also added another kind of measure, called CompositeMeasure, which is leveraged whenever you have integration on adapted meshes. Not sure if this has to be taken into account when updating this branch from master ...

Forget this. I now see you are already taking into account `CompositeMeasure` in your PR ...

amartinhuertas:
Also dont forget to add the corresponding entry in NEWS.md ... Thanks!

ericneiva:
Hi, @amartinhuertas,

> I agree the API is much cleaner now.

Good to know :)

> Can we may be add the old signatures and add corresponding deprecated warnings so that we avoid a major release?

Done. Let me know if the deprecated warnings are ok.

> Also dont forget to add the corresponding entry in NEWS.md

Done.

Thank you very much for your review!

amartinhuertas:
FYI, just in case you did not know, I found this article on deprecation quite useful: 

https://invenia.github.io/blog/2022/06/17/deprecating-in-julia/

All good now. Accepting PR!

ericneiva:
Great, thanks!

> FYI, just in case you did not know, I found this article on deprecation quite useful:

Useful, indeed, thanks for sharing, @amartinhuertas

PR #893, opened 2023-03-30
Remove duplicated test

kristianmeyer:
Hi,

Just a small change, since I found that this test was duplicated when I went over the tests.

Best regards
Kristian


amartinhuertas:
Hi @kristianmeyer, thanks for the PR, but please note that scaling from the left and from the right, although mathematically equivalent, it is not from a programming point of view. In particular, when dealing with overloaded methods in programming languages both scalings do not trigger the same method. Thus, I think this is not actually a duplicated test.

PR #902, opened 2023-05-13
deleting nede_tet_1.vtu generated during NedelecRefFEs tests

kishore-nori:
The "`nede_tet_1.vtu`" file is generated from the `NedelecRefFEs` tests, I am adding a line to delete it, as overlooking it after performing global tests on local machine could add it to the git branch. I added it outside `NedelecRefFEsTests.jl` file so that testing only this particular module, can allow us to inspect the "`nede_tet_1.vtu`" (which I think is its purpose?) rather than automatically deleting it (of course, in that case we have to delete it ourselves, but it is inconvenient while performing global tests).


amartinhuertas:
I think the usual approach around this consists on the usage of `d = mktempdir()` + `rm(d,recursive=true)`

Search for use cases in the current Gridap source code. 

I dont think the vtk case is checked for consistency. The test is just ensuring that the code does not crash.

Thus we can follow the aforementioned approach.

Issue #905, opened 2023-05-25
gmsh mesh and CellState projections


christopherfear:
I am looking to project a CellState from the triangulation of a quad mesh created in gmsh.

Triangles created in gmsh/ a domain of quads made in Gridap (with the domain and partition functions) work with no issues.

The MWE of the code is as follows:

`using Gridap
using LinearAlgebra
using GridapGmsh

model = GmshDiscreteModel("C:/Users/ttcf5/gui.msh")

order = 1
reffe = ReferenceFE(lagrangian,Float64,order)
V = TestFESpace(model,reffe)
U = TrialFESpace(V)

degree = 2*order
Ω = Triangulation(model)
dΩ = Measure(Ω,degree)

f = CellState(1.0, dΩ)

function project(q,model,dΩ,order)
    reffe = ReferenceFE(lagrangian,Float64,order)
    V = FESpace(model,reffe,conformity=:L2)
    a(u,v) = ∫( u*v )*dΩ
    l(v) = ∫( v*q )*dΩ
    op = AffineFEOperator(a,l,V,V)
    qh = solve(op)
    qh
end

fh = project(f, model, dΩ, order)`

The error is:

> ERROR: 
It is not possible to evaluate the given CellState on the given CellPoint.
a CellState can only be evaluated at the CellPoint it was created from.
If you want to evaluate at another location, you would need first to project the CellState
to a FESpace (e.g. via a L2 projection).

Am I misunderstanding something?
Could anyone help me to solve this problem?

kishore-nori:
Hi, 

What you have should work (just to make sure, I ran and checked, but on a different model, but that shouldn't matter). However, it might have so happened you re-ran `dΩ = Measure(Ω,degree)` but did not re-run the `f = CellState(1.0, dΩ)` once `dΩ` was updated and this causes a problem, as although the new `dΩ` is effectively the same but not identical (i.e. it is not the same object anymore, and doesn't satisfy `===`, hence it results in the error you bumped into). You can restart the `julia` environment and re-run everything sequentially, and it should work. Having said that, a safer approach would be have `f = CellState(1.0, dΩ)` inside your `project` function (better to also have `degree = 2*order` and `dΩ = Measure(Ω,degree)`), which will ensure `f` to be consistent with `dΩ` and avoids the earlier discrepancy.

christopherfear:
Hi,

[...]

PR #906, opened 2023-05-30
changing map to lazy_map to fix issue #905

kishore-nori:
Changing `map` to `lazy_map` to fix issue #905. This is done to return identical `CellPoint` objects. Tests on my local machine have passed. 

The issue was `get_cell_point(dΩ)`, when `Measure` `dΩ` is built from a `GmshDiscreteModel` based on `QUAD`s, returns `CellPoint` objects which are non-identical i.e. (`get_cell_points(dΩ) === get_cell_points(dΩ)` is not satisfied). This is due to non-identical `cell_phys_point` field in `CellPoint`. This results in the issue #905, where `CellState` constructed based on `dΩ` doesn't satisfy the compatibility test during integration. 

Haven't included a test, as I ll then have to include the `.msh` file as well, or may be I can add a test to `GridapGmsh.jl` pakcage, which ever seems reasonable..


amartinhuertas:
> (get_cell_points(dΩ) === get_cell_points(dΩ) is not satisfied).

How can it be that this condition is not satistied? Isn't dΩ the same object in your comment?

kishore-nori:
> > Hav't included a test, as I ll then have to include the .msh file as well, or may be I can add a test to GridapGmsh.jl pakcage, which ever seems reasonable..
> 
> Can the issue be reproduced with a manually built `UnstructuredDiscreteModel` of quadrilaterals? see, e.g., for an example snippet [gridap/GridapP4est.jl#21 (comment)](https://github.com/gridap/GridapP4est.jl/issues/21#issuecomment-1384635611)

I tried building `UnstructuredDiscreteModel` on top of `CartesiandDiscreteModel` but it doesn't fail in that case. I ll try out manual building and get back

kishore-nori:
> > (get_cell_points(dΩ) === get_cell_points(dΩ) is not satisfied).
> 
> How can it be that this condition is not satistied? Isn't dΩ the same object in your comment?

No.. that is exactly the source of error for this issue, and this PR fixes it. Even if `dΩ` is the same, `get_cell_points(dΩ) === get_cell_points(dΩ) ` don't end up same.. due to non identical `cell_phys_point` which somewhere deep inside uses `map` instead of `lazy_map`, this is only peculiar to Quad mesh with `GmshDiscreteModel`

amartinhuertas:
> I tried building UnstructuredDiscreteModel on top of CartesiandDiscreteModel but it doesn't fail in that case. I ll try out manual building and get back

Don't use the example provided verbatim. In the example, I am generating a non-conforming mesh, which is currently not well supported by Gridap. You can, e.g., extract the submesh corresponding to the first four cells in the example

kishore-nori:
>see, e.g., for an example snippet https://github.com/gridap/GridapP4est.jl/issues/21#issuecomment-1384635611

Thank you for the pointer.. now the issue is reproducible :) and I also checked that the PR fixes it. See following, also including the issue #905 example, where should I put the below test? `GridapTests`?

```julia
using Gridap
using GridapGmsh
using FillArrays

ptr  = [ 1, 5, 9, 13, 17 ]
data = [ 1,2,4,5, 2,3,5,6, 4,5,7,8, 5,6,8,9 ]
cell_vertex_lids = Gridap.Arrays.Table(data,ptr)
node_coordinates = Vector{Point{2,Float64}}(undef,11)

for j in 1:3
  for i in 1:3
    node_coordinates[3*(j-1)+i]=Point{2,Float64}((i-1)/2.0,(j-1)/2.0)
  end
end

polytope = QUAD
scalar_reffe = Gridap.ReferenceFEs.ReferenceFE(polytope,Gridap.ReferenceFEs.lagrangian,Float64,1)
cell_types = collect(Fill(1,length(cell_vertex_lids)))
cell_reffes = [scalar_reffe]
grid = Gridap.Geometry.UnstructuredGrid(node_coordinates,
                                        cell_vertex_lids,
                                        cell_reffes,
                                        cell_types,
                                        Gridap.Geometry.NonOriented())
model = Gridap.Geometry.UnstructuredDiscreteModel(grid)

Ω = Triangulation(model)
dΩ = Measure(Ω,2)

# the problem is below:
dΩ.quad.cell_point === dΩ.quad.cell_point # true
dΩ.quad.cell_point === get_cell_points(dΩ) # false
get_cell_points(dΩ) === get_cell_points(dΩ) # false, the source of the issue #905

# issue #905 example

order = 1
function project(order,Ω)
  degree = 2*order
  dΩ = Measure(Ω,degree)
  # f(x) = sin(norm(x))
  f = CellState(1.0,dΩ)
  reffe = ReferenceFE(lagrangian,Float64,order)
  V = FESpace(model,reffe,conformity=:L2)
  a(u,v) = ∫( u*v )dΩ
  l(v) = ∫( v*f )*dΩ
  op = AffineFEOperator(a,l,V,V)
  fh = solve(op)
  fh
end

fh = project(order,Ω) # fails

ERROR:

It is not possible to evaluate the given CellState on the given CellPoint.

a CellState can only be evaluated at the CellPoint it was created from.
If you want to evaluate at another location, you would need first to project the CellState
to a FESpace (e.g. via a L2 projection).
```

amartinhuertas:
> Where should I put the below test? GridapTests?

Yes. Let us put it in that directory. `issue_905.jl` file.

PR #921, opened 2023-07-18
Added a test to RefinementRulesTests.jl that fails and should not

amartinhuertas:
Solves failing test https://github.com/gridap/Gridap.jl/blob/09b6e9b6fb8588d5087a87054f471ee14559ae3f/test/AdaptivityTests/RefinementRulesTests.jl#L26

### DIFF
--- a/test/AdaptivityTests/RefinementRulesTests.jl
+++ b/test/AdaptivityTests/RefinementRulesTests.jl
@@ -22,6 +22,9 @@ for poly in polys
   rr_red = Adaptivity.RedRefinementRule(poly)
   Adaptivity.test_refinement_rule(rr_red)
 
+  rrules=[rr_generic, rr_white, rr_red]
+  @test isconcretetype(eltype(rrules))
+
   n_edges = num_faces(poly,1)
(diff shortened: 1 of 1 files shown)


JordiManyer:
@amartinhuertas Yes, this is something I already encountered in the past. We discussed it, and we ended up agreeing that we will probably have to remove most template parameters from the `RefinementRule` structure in order to not have problems with mixed meshes... It's a shame, but it should not lead to performance issues since the number of different refinement rules should stay relatively small at all times.

JordiManyer (approved):

amartinhuertas:
> We discussed it, and we ended up agreeing that we will probably have to remove most template parameters from the RefinementRule structure in order to not have problems with mixed meshes...

Yes, we discussed it, and took some action, as far as I remember (namely, redefinition of type parameters for `RefinementRule`). These changes solved issues with TRIs but not with QUADs, even with no mixed meshes. In any case, as you say, with mixed meshes is going to be hard to bypass this problem if we stick to type paremeters in RefinementRule.

PR #934, opened 2023-08-15
add real/imag for VectorValue

HelgeGehring:
Thanks for the great project!

I recently ran into an error while exporting complex VectorValues.
This small PR allows to apply `real`/`imag` on VectorValues


HelgeGehring:
@fverdugo as discussed, I added some tests, could you have a look if it's fine like this?

PR #954, opened 2023-11-15
Nonlinear Runge-Kutta

oriolcg:
Updated `RungeKutta.jl` to enable nonlinear operators following the same style as in `EXRungeKutta.jl`.


santiagobadia:
Thanks @oriolcg

PR #956, opened 2023-11-21
Bugfix: distributed `add_tag_from_tags!` not working

JordiManyer:
`add_tag_from_tags!` not working for distributed models created from a single serial model in debug mode. This was caused by arrays being aliased. Fixed by copying the arrays.


amartinhuertas:
Question: Do we need to solve this in Gridap.jl? In other words, can be instead solve the issue at the place where it is actually needed?

JordiManyer:
> Question: Do we need to solve this in Gridap.jl? In other words, can be instead solve the issue at the place where it is actually needed?

After discussion, we agreed this is actually also good for the serial case. Different model portions should have independent face labelings. Also, the arrays that get copied are tiny.

PR #990, opened 2024-04-05
Fix for issue #905, replicating the PR #906 

kishore-nori:
Due to some rebasing mess up from my side, I am replicating the PR (https://github.com/gridap/Gridap.jl/pull/906, which I closed to clean up the history). The changes in this PR and related tests, fix and isolate the issue #905.

cc @amartinhuertas


kishore-nori:
I just realised that this PR is related to the recently merged PR https://github.com/gridap/Gridap.jl/pull/987. The approaches to deal with the issue are different PR https://github.com/gridap/Gridap.jl/pull/987, makes things work for equivalent `CellPoint`s and this one fixes the problem that `get_cell_points` generates a different object (which probably is not good?) by returning identical `CellPoint` objects, thereby fixing the cause of the issue, which breaks 

```julia
function evaluate!(cache,f::CellState,x::CellPoint)
  if f.points === x
    f.values
  else
    @unreachable """\n
    It is not possible to evaluate the given CellState on the given CellPoint.

    a CellState can only be evaluated at the CellPoint it was created from.
    If you want to evaluate at another location, you would need first to project the CellState
    to a FESpace (e.g. via a L2 projection).
    """
  end
end
```

changed to `f.points == x` in PR https://github.com/gridap/Gridap.jl/pull/987, I feel both ways are required..

PR #991, opened 2024-04-08
Added `axpy_entries` for `AbstractBlockMatrix`

oriolcg:
Added `axpy_entries!` function for `AbstractBlockMatrix`. Without this function specialization there is a bottleneck in the solution of transient problems with block matrices.


JordiManyer:
Hi @oriolcg , thanks for reporting this! I'll merge it into the release branch when the tests pass.

PR #1000, opened 2024-04-17
Matrix assembly rework

JordiManyer:
This relates to what we discussed. @amartinhuertas @fverdugo 

A little context: 

The current implementation of the sparse matrix assembly is (loosely) done as follows:
  - when first assembled, we assemble entries `InserterCSC` object, with a similar layout to `SparseMatrixCSC`. 
  - when re-assembled, we assemble entries directly onto the sparse matrix.
Both of these have the advantage of saving memory, but at the cost of doing sparse row/col index searches every time we assemble an entry. 

We would like to investigate the option of working with COO-type structures. These are more expensive memory-wise, but can save quite a lot of time when re-assembling the matrix. 

I've added a MWE on a possible implementation of the latter. Also, here are some results from the benchmarks at the end: 
  - When first assembling: Both approaches take the same time, but COO has 3x the memory footprint. 
  - When reassembling: COO is twice as fast than the current implementation.


oriolcg:
Hi @JordiManyer , what is the status of this PR? Have you tried the new implementation for Nonlinear/transient problems?

JordiManyer:
@oriolcg I don't have time atm, I'll try to push this further after the summer. I am not focusing complicated problems yet, but obviously anything that assembles repeatedly (like NL or transient) would benefit from this. I am worried about the memory footprint though... I think I can make most of it lazy so that we don't have to keep any Float values in memory, but it make take some time to work it out.

PR #1004, opened 2024-04-30
use OrderedDict

### LINKED ISSUE #1002: non-deterministic matrix assembly for multiple domains
@mochen4 pointed out via email that Gridap's matrix assembly yields non-deterministic roundoff errors, e.g. the matrices from [this tutorial](https://gridap.github.io/Tutorials/dev/pages/t018_TopOptEMFocus/) change on the order of machine precision from run to run.   This makes debugging harder, and is especially inconvenient if the Gridap outputs are fed into some calculation exhibiting chaos or instability (e.g. we noticed it in optimization problems exhibiting spontaneous symmetry breaking, which converge to a local optimum that is sensitively affected by tiny discrepancies).

@fverdugo wrote to explain:

> we use a `Dict` to store the contributions of several domains to an integral. This can definitely be a cause of operations being done in non deterministic order.

and indeed the issue arises in problems involving multiple domains.   Essentially, `Dict` iteration depends on the hashing algorithm, and Julia's [default hashing algorithm](https://github.com/JuliaLang/julia/blob/8423426213f58c786de33b10ab41d90fe89afe04/base/hashing.jl#L38) for a `struct` uses the pointer address (`objectid`), which is effectively non-deterministic.   Changing iteration order can change the floating-point roundoff errors of a sum.

The solution is, in principle, simple — just implement a custom `hash` function (and `==`) that is deterministic for any object that is used as a `Dict` key.     Question: which data structures in Gridap need this?

mochen4:
Fixes #1002. Use OrderedDict to ensure deterministic behavior.

### DIFF
--- a/src/CellData/DomainContributions.jl
+++ b/src/CellData/DomainContributions.jl
@@ -2,7 +2,6 @@ using DataStructures
 
 """
 """
-
 struct DomainContribution <: GridapType
   dict::OrderedDict{Triangulation,AbstractArray} # ordered so that iteration is deterministic (#1002)
 end
--- a/src/CellData/DomainContributions.jl
+++ b/src/CellData/DomainContributions.jl
@@ -1,7 +1,6 @@
-
-"""
-"""
 using DataStructures
+"""
+"""
 
 struct DomainContribution <: GridapType
   dict::OrderedDict{Triangulation,AbstractArray} # ordered so that iteration is deterministic (#1002)
--- a/src/CellData/DomainContributions.jl
+++ b/src/CellData/DomainContributions.jl
@@ -4,7 +4,7 @@
 using DataStructures
 
 struct DomainContribution <: GridapType
-  dict::OrderedDict{Triangulation,AbstractArray}
+  dict::OrderedDict{Triangulation,AbstractArray} # ordered so that iteration is deterministic (#1002)
 end
 
 DomainContribution() = DomainContribution(IdDict{Triangulation,AbstractArray}())
--- a/src/CellData/DomainContributions.jl
+++ b/src/CellData/DomainContributions.jl
@@ -1,8 +1,10 @@
 
 """
 """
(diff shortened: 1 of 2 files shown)


src/CellData/DomainContributions.jl
```diff
@@ -1,8 +1,10 @@
 
 """
 """
+using DataStructures
```
stevengj, line 4: You need to move this up, above the docstring (which is currently empty).  Just make the `using` statement the first line of the file.

src/CellData/DomainContributions.jl
```diff
@@ -1,8 +1,10 @@
...
+
 struct DomainContribution <: GridapType
-  dict::IdDict{Triangulation,AbstractArray}
+  dict::OrderedDict{Triangulation,AbstractArray}
```
stevengj, line 7:
```suggestion
  dict::OrderedDict{Triangulation,AbstractArray} # ordered so that iteration is deterministic (#1002)
```

src/CellData/DomainContributions.jl
```diff
@@ -1,8 +1,9 @@
-
+using DataStructures
```
stevengj, line 1:
```suggestion
using DataStructures

```

src/CellData/DomainContributions.jl
```diff
@@ -1,8 +1,10 @@
...
 
 """
 """
+
```
stevengj, line 5:
```suggestion
```
No blank line here.  This is supposed to be attaching a [docstring](https://docs.julialang.org/en/v1/manual/documentation/) to the `struct DomainContribution` type.  (The docstring is currently empty as a placeholder, I guess.)

stevengj (approved):

amartinhuertas:
Can you please add an entry to the news.md file?? Thanks!

PR #1009, opened 2024-05-22
Fix incorrect dof value type for linear constraints fe spaces

dnuy:
When e.g. solving the Helmholtz equation we set `vector_type=Vector{ComplexF64}` for the `FESpace`, in that case we obtain correctly that `get_dof_value_type(V) == ComplexF64`.
When the space is a `FESpaceWithLinearConstraints` then `get_dof_value_type` incorrectly returns `Float64`; based on the default implementation in `FESpaceInterface.jl`.

My fix is to add
```Julia
get_dof_value_type(f::FESpaceWithLinearConstraints) = eltype(get_vector_type(f.space))
```
to `FESpacesWithLinearConstraints.jl`.

PS: It might be that the default implementation for `get_dof_value_type` in `FESpaceInterface.jl` is incorrect, as also for `V` a `FESpace` with `vector_type=Vector{ComplexF64}` I obtain `get_dof_value_type(get_fe_basis(V), get_fe_dof_basis(V)) == Float64` instead of `ComplexF64`.

I have extended the tests in `FESpacesWithLinearConstraintsTests.jl` to have `vector_type=Vector{ComplexF64}` and added appropriate tests. The first commit in this merge request just shows the current bug with `@test get_dof_value_type(Vc) == ComplexF64 broken=true` and then fixes the bug as above and runs the test again without the `broken=true`; the second commit corrects the bug in `FESpacesWithLinearConstraints.jl` and just leaves a `@test get_dof_value_type(Vc) == ComplexF64` in place.

### DIFF
--- a/src/FESpaces/FESpacesWithLinearConstraints.jl
+++ b/src/FESpaces/FESpacesWithLinearConstraints.jl
@@ -470,6 +470,10 @@ function get_vector_type(f::FESpaceWithLinearConstraints)
   get_vector_type(f.space)
 end
 
+function get_dof_value_type(f::FESpaceWithLinearConstraints)
+  eltype(get_vector_type(f.space))
+end
+
 function get_fe_basis(f::FESpaceWithLinearConstraints)
   get_fe_basis(f.space)
 end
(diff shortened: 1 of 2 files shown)


JordiManyer:
Should I close this then?

dnuy:
Yes I think this can be marked as resolved. Thanks.
What could I have done to have had this taken care of more quickly, as my fix was exactly the same as the one which was accepted?

JordiManyer:
@dnuy I just missed it, sorry. I try to be on top of these things, but the team is quite busy. I guess next time just tag me directly if noone responds within the first couple days. My apologies.

PR #1014, opened 2024-07-12
Performance mode

JordiManyer:
Hi everyone! 

For some context: 

The Julia flag `--check-bounds=no` is, as far as I understand, currently unusable due to it causing type instabilities within the compiler. Within Gridap, we use this option to enable/disable the `@check` macro. This means we do not currently have an option to deactivate such checks. Some heavy calculations are hidden behind those checks, so it would be nice to have  the option to deactivate them.

Moreover, there are currently packages that statically (at compile time) address this issue. In particular, the `Preferences.jl` package offers an unified approach to deal with this. Very prominent Julia packages have been moving towards this approach, for instance `MPI.jl` through `MPIPreferences.jl`. 

Given these two last points, I propose the following PR. It introduces a constant variable `execution_mode` that can be set to either `debug` (default) or `performance`. The latter deactivates the `@check` macro at compile time. 

As a side note, I think we should also move towards this approach within `GridapPETSc.jl` and `GridapP4est.jl`, instead of depending on environment variables (again, the approach taken by `MPI.jl`). 

I would appreciate some feedback on the issue, to make sure we are all on the same page. 
Thanks!


amartinhuertas:
> The Julia flag --check-bounds=no is, as far as I understand, currently unusable due to it causing type instabilities within the compiler. 

Can you please share a reference on this? For my understanding ... Thanks!

JordiManyer:
> > The Julia flag --check-bounds=no is, as far as I understand, currently unusable due to it causing type instabilities within the compiler.
> 
> Can you please share a reference on this? For my understanding ... Thanks!

https://github.com/JuliaLang/julia/issues/48245

amartinhuertas:
For the records, the PR is mostly based on the following doc page https://juliapackaging.github.io/Preferences.jl/stable/#API

fverdugo:
I like the switch based on Preferences.jl, which provably makes sense even beyond the issue with `--check-bounds`.

But this is only a solution to part of the problem unfortunately. 

In Gridap (and multitude of Julia packages) there are a lot of loops without `@inbounds`, which will have bounds checks if Julia removes `--check-bounds=no` .... I have been reading the issue https://github.com/JuliaLang/julia/issues/48245  and I don't find any clear solution at the moment apart from populating the packages with `@inbounds`.

Do you have any idea/opinion on how to proceed regarding array indexing if julia removes `--check-bounds`?

fverdugo (approved):
See my comment above.

JordiManyer:
> In Gridap (and multitude of Julia packages) there are a lot of loops without `@inbounds`, which will have bounds checks if Julia removes `--check-bounds=no` .... I have been reading the issue [JuliaLang/julia#48245](https://github.com/JuliaLang/julia/issues/48245) and I don't find any clear solution at the moment apart from populating the packages with `@inbounds`.
> 
> Do you have any idea/opinion on how to proceed regarding array indexing if julia removes `--check-bounds`?

Following Tim Besard's comments on that issue, I believe the GPU people have just gone ahead and created their own compiler (which we obviously cannot do...). From what I understand, however, the performance should be better? 

Another thing coming with Julia 1.11 is the new `Memory` type (see [this](https://pretalx.com/juliacon2024/talk/CYJ7V7/)) which is supposed to be even lower-level than `Array`. The idea (I think) is to reduce some of the overhead in the case where all the fancy stuff from `Array` is not needed. I do not know if one of the things it removes is bounds-checking, but we could look into it.

PR #1015, opened 2024-07-22
Passing kwargs to simplexify refinement

pmartorell:
@santiagobadia @JordiManyer 

This PR passes the kwargs of refine to the simplexify core functions. In particular, it is tested for `positive=true`, which generates simplexified partitions with positive volume cells.


JordiManyer:
Hey @pmartorell ! Thank you for this! I did not know simplexify had any kwargs, but it is a good thing to have.

PR #1018, opened 2024-08-02
Add coarsening in 2D for newest vertex bisection

aerappa:
@JordiManyer, @amartinhuertas :

 @wangzuodong1997 and I have been working on implementing coarsening in the case of 2D refinement.  Specifically, we have only implemented modifying the `Topology` directly without bothering with the infrastructure of `rrules` etc. currently implemented for refinement that (if I understand correctly) allow interpolation and projection between parent and children meshes.

Our refinement algorithm closely resembles the one [here](https://aam.uni-freiburg.de/mitarb/ehemalige/schoen/publications/BarSch12-pre.pdf) and has the attractive feature that it doesn't not require any mesh history to perform local coarsening that preserves mesh continuity. At a high level, the user selects a set of cells to coarsen, and our algorithm identifies if any of these cells contain a "good node" which can be removed with some of it's edges without generating any hanging nodes.

From your point of view, what are the next steps? Can `AdaptivityGlue` be reused for coarsening? It seems rather specific to refinement at the moment.

Please find a small example below demonstrating the coarsening.

Before:
before
After:
after

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Low level optimisations to reduce allocations. `AffineMap` renamed to `AffineField`. New `AffineMap <: Map`, doing the same as `AffineField` without struct allocation. New `ConstantMap <: Map`, doing the same as `ConstantField` without struct allocation. Since PR[#1043](https://github.com/gridap/Gridap.jl/pull/1043).
+- `ConstantFESpaces` can now be built on triangulations. Since PR[#1069](https://github.com/gridap/Gridap.jl/pull/1069)
 
 ## [0.18.8] - 2024-12-2
 
--- a/src/FESpaces/ConstantFESpaces.jl
+++ b/src/FESpaces/ConstantFESpaces.jl
@@ -1,49 +1,63 @@
 
 """
-struct ConstantFESpace <: SingleFieldFESpace
-  # private fields
-end
+    struct ConstantFESpace <: SingleFieldFESpace
+      # private fields
+    end
+
+    ConstantFESpace(model::DiscreteModel; vector_type=Vector{Float64}, field_type=Float64)
+    ConstantFESpace(trian::Triangulation; vector_type=Vector{Float64}, field_type=Float64)
+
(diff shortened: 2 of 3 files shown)


JordiManyer:
@aerappa Thanks for the contribution! 

> From your point of view, what are the next steps? Can AdaptivityGlue be reused for coarsening? It seems rather specific to refinement at the moment.

We are currently already using what we have for coarsening in GridapP4est, so nothing new needs to be added. The mixed glue should be ok for mixed refining/coarseningm (i.e the adapt function). If you only have coarsening, the refinement glue does the job just fine (since coarsening is basically refining in reverse). Simply put the coarsened model as the parent, not the child.

JordiManyer:
Hey @aerappa @wangzuodong1997 , any updates on this? I recently merged some stuff for AMR that might also interest you guys, you could have a look at the tests I created in `test/Adaptivity`.  Let me know if I can be of help.

wangzuodong1997:
> Hey @aerappa @wangzuodong1997 , any updates on this? I recently merged some stuff for AMR that might also interest you guys, you could have a look at the tests I created in `test/Adaptivity`. Let me know if I can be of help.

Hi. Yes, the added example is helpful, thanks :) 
Here is some progress:
1. I tried with @aerappa to fix the Glue issue, but we failed.
2. Instead, we implemented the newest vertex bisection scheme and the corresponding coarsening scheme which work well. I tested these schemes with heat equation with moving singularity, everything looks fine.
3. However, we did not store the hierarchy which makes the current schemes slow. 

@aerappa can implement the hierarchy and probably merge these schemes in  Gridap, how do you think? @JordiManyer

JordiManyer:
I would start by pulling from `master` since this branch is quite far behind at the moment. Once that is done, I can maybe have a look at the code and see how you guys can implement the glue.

wangzuodong1997:
> I would start by pulling from `master` since this branch is quite far behind at the moment. Once that is done, I can maybe have a look at the code and see how you guys can implement the glue.

Thanks! I did not yet push my refine/coarsen code in git. Please let me know when I can push the code :)

JordiManyer:
@wangzuodong1997 I do not follow. It's @aerappa 's fork, so its fully under his control. Pulling from master is something anyone can do as well, no need for special permissions.

PR #1034, opened 2024-10-08
Fix broken GMSH link in Readme

ThePurox:


amartinhuertas:
Thanks for the typo fix @ThePurox!

PR #1038, opened 2024-10-15
MultiValues component names export

### LINKED ISSUE #1036: MultiValues components naming in vtk exports
Currently, we don't name the `MultiValue` components in the vtk exports, which may be confusing as Paraview automatically rename them in function of the number of components. So a 2D symmetric tensors has the components named like 3D vector (X Y Z), and Paraview assumes Voigt ordering  (XX, YY, ...) for 3D symmetric tensor while we export Cartesian like ordering (XX, XY, ...).

I propose to explicitly write the component names of every MultiValue in the exported vtk files to prevent Paraview's auto-naming.
Also, I propose to label the component X, Y and Z for dimension d<=3 similarly to what Paraview currently does, and 1,2 ... d for larger dimensions to keep generality.

I implementing this for serial export in this [branch](https://github.com/Antoinemarteau/Gridap.jl/tree/vtk_componant_names_export) but currently WriteVTK.jl doesn't allow writing the component names in pvtk [153](https://github.com/JuliaVTK/WriteVTK.jl/issues/153#issue-2584430777), so wait before opening PR.

Is that alright ?

Antoinemarteau:
Fixes #1036.

### DIFF
--- a/src/Visualization/Vtk.jl
+++ b/src/Visualization/Vtk.jl
@@ -94,10 +94,12 @@ function create_vtk_file(
 
   if num_cells(trian)>0
     for (k,v) in celldata
-      vtk_cell_data(vtkfile, _prepare_data(v), k; component_names=_data_component_names(v))
+      comp_names = _data_component_names(v)
+      vtk_cell_data(vtkfile, _prepare_data(v), k; component_names=comp_names)
     end
     for (k,v) in nodaldata
-      vtk_point_data(vtkfile, _prepare_data(v), k; component_names=_data_component_names(v))
+      comp_names = _data_component_names(v)
+      vtk_point_data(vtkfile, _prepare_data(v), k; component_names=comp_names)
     end
   end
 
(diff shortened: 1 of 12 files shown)


src/Visualization/Vtk.jl
```diff
@@ -94,10 +94,12 @@ function create_vtk_file(
 
   if num_cells(trian)>0
     for (k,v) in celldata
-      vtk_cell_data(vtkfile, _prepare_data(v), k)
+      comp_names = _data_component_names(v)
```
JordiManyer, line 97: I would replace this by
```julia
component_names = _data_component_names(v)
vtk_cell_data(vtkfile, _prepare_data(v), k; component_names)
```
and idem everywhere else.

JordiManyer (commented):
I've added a very minor comment. 
Otherwise, everything looks great. Remember to pull from master and update `NEWS.md`. After you are done I'll accept the PR.

src/Visualization/Vtk.jl
Antoinemarteau: Oh I didn't know this trick, I changed these.

Antoinemarteau: Actually the setindex overload is so cursed that I cannot avoid repeating the kwarg name in it :') .

PR #1048, opened 2024-11-08
Moment based reffes

JordiManyer:
So we've been having a look at the current implementation of Raviart-Thomas and Nedelec. It has become clear that the current implementation

- has issues with high-order, that have to do with how we choose our basis/prebasis when computing the moments. 
- has issues related to scaling in mixed formulations. 
- is outright wrong when using high-order approximations for the geometry. 
- is very hard to read, modify and extend. 

Moreover, there has been interest lately in implementing more types of moment-based discretisations like Crouzeix-Raviart. 

We believe the best way to solve all these issues at once is to create some extendable machinery to easily create moment-based FEs. 

## TODOs: 

- [ ] Create the moment-based machinery 
- [ ] Create the pullback machinery
- [ ] Replicate RT and ND in 2D/3D with structured meshes
- [ ] Replicate RT and ND in 2D/3D with unstructured meshes
- [ ] Benchmark against current implementations

## New FEs: 

- [ ] Implement Crouzeix-Raviart

## Other: Polynomial bases

The solution to the first issue requires extending Jacobi-type polynomial bases. Although it could be done blindly, I think we could try to merge a lot of code together. In the future we might want to have even more basis of polynomials, which would require a lot of copied code. 

- [ ] Extend Monomial bases to accomodate Jacobi and other types of polynomials
- [ ] Benchmark extensively


src/ReferenceFEs/ReferenceFEs.jl
```diff
@@ -197,6 +196,7 @@ export bdm
 export nedelec
 export bezier
 export modalC0
+export cr
```
Antoinemarteau, line 199: shouldn't we name Crouziex-Raviart reffename: crouziex_raviart = CrouziexRaviart() for clarity and consistency with raviart_thomas = RaviartThomas() ?

Jai-Tushar: In the literature, people just refer to it as "cr" just like "bdm". I would say that we should name "raviart_thomas" as "RT" instead, which is how people refer to it in the literature and probably modify "cr" to "CR" and "bdm" to "BDM".

JordiManyer: I prefer longer names

JordiManyer: Also: If you don't know what Raviart-Thomas is, you can google it. Going the other direction is not as easy.

JordiManyer:
@Antoinemarteau  we need to reproduce/have a look at #1222

src/ReferenceFEs/MomentBasedReferenceFEs.jl
```diff
@@ -0,0 +1,543 @@
...
+
+  if change_dof
+    # Use geometric decomposition to avoid shapefuns change of basis.
+    # The dofs are computed as dual to the shapefuns, using a change of basis
+    # from the moment basis
+
+    @assert has_geometric_decomposition(prebasis, p, conformity)
+    shapefuns, predofs = prebasis, dof_basis
+
+    # TODO: This is very ugly.... do we really need to be adding stuff here every time
```
JordiManyer: This is sketchy... As we add more stuff, these conditions will grow?

Issue #875, opened 2023-02-12
Laplacian of product of `CellField`s doesn't work. 


kishore-nori:
The Laplacian of product of `CellField`s (formed using the same `FESpace` and `Triangulation`) fails. I came across the following,
https://github.com/gridap/Gridap.jl/blob/672e3d9027246bfa677c07b4e9fabb9e1fde9eb4/src/Fields/FieldsInterfaces.jl#L94-L102
But the error doesn't come from here, the evaluation doesn't hit the above, at least from preliminary debugging observation. The following is the MWE:

```julia
using Gridap
using Test

domain = (0.,1.,0.,1.)
partition = (3,3)
model = CartesianDiscreteModel(domain,partition)

g(x) = exp(-norm(x))
reffe = ReferenceFE(lagrangian,Float64,2)
V = TestFESpace(model,reffe,conformity=:H1,dirichlet_tags="boundary")
U = TrialFESpace(V,g)

n = num_free_dofs(U)
uh = FEFunction(U,rand(n))
qh = FEFunction(U,rand(n))

p = Point(0.1,0.2)

cf1 = ∇(uh + qh)
@show @test cf1(p) ≈ ∇(uh)(p) + ∇(qh)(p) # works 

cf2 = ∇(uh*qh)
@show @test cf2(p) ≈ (qh*∇(uh))(p) + (uh*∇(qh))(p) # works 

cf3 = Δ(uh + qh) # equiv to tr(∇∇(⋅)), making it an OperationCellField
@show @test cf3(p) ≈ Δ(uh)(p) + Δ(qh)(p) # works 

cf4 = ∇∇(uh + qh)
@show @test cf4(p) ≈ ∇∇(uh)(p) + ∇∇(qh)(p) # works 

# Although the CellField is built, when evaluated it results in error
cf5 = ∇∇(uh*qh) # works 
cf5(p) # fails 

cf6 = Δ(uh*qh) # even the construction fails 
```
I would be happy to fix the related issue and implement the second order derivatives based on your guidance and suggestions, as the `@notimplemented` message directs. 

cc @amartinhuertas, @oriolcg (I think latest commits along these lines where your contributions)

carlodev:
I don't know if this is completely correct. I looked also as `push_∇(∇a::Field,ϕ::Field) = pinvJt(∇(ϕ))⋅∇a`:
```
function push_∇∇(∇∇a::Field,ϕ::Field)
  s = pinvJt(∇(ϕ))
[...]

Issue #69, opened 2019-08-06
Implement second order derivatives (Implement Laplacian)


fverdugo:
Required for VMS, Explicit error estimators etc

PR #1053, opened 2024-11-17
Adding rules and methods for hessian of sum of scalar CellFields and product of two scalar CellFields

kishore-nori:
- added rules and methods for hessian of sum of scalar Fields (hence CellFields)
- added rules and methods for hessian of product of two scalar Fields (hence CellFields)
- added corresponding tests to check the new functionality

this closes issue https://github.com/gridap/Gridap.jl/issues/875


JordiManyer:
Hey @kishore-nori ! I need this actually, can we merge it? I think we would only need to pull from master and increase a bit the number of tests.

JordiManyer:
Also closes #69

kishore-nori:
Hi @JordiManyer I have added few more tests, and it improves the codecov but seem like haven't been able to trigger the code additions in `src/Fields/ApplyOptimizations.jl`. Having said that, it works on `CellField`s and `FEFunction`s so should be alright.

PR #1077, opened 2025-02-06
Handeling points outside domain

lisavdlinde:


JordiManyer (changes requested):
Hi @lisavdlinde , thanks for the contribution! I do have some comments on the changes though.

- I've added some comments to the code. 
- Also, please add `@notimplementedif searchmethod.accept_points_outside` to all functions that potentially allow the user to use the kwarg but which do not take it into consideration. For instance, the evaluation with an array of points.

JordiManyer:
On top of this: 
- Please pull from master, I just merged some other stuff. 
- Please update the `NEWS.md` file with the changes
Thanks!

Issue #1075, opened 2025-01-15
GridapODEs - Changes in the LinearStageOperator Struct 


shreyas02:
Hi all, 

I am working on implementing strong form coupling for **linear** transient fluid-structure interaction problems. For that, I need to modify the linear system before solving the matrix for each time step. I want to implement this with as few changes in the Gridap library as possible. We need previous time step data and some ODE-specific data for these modifications. 

The solution we arrived at was creating a `CustomSolver` wrapper (`solve`, `solve!`, `symbolicsetup`,  `numericalsetup`) and performing all the matrix modifications in the custom `numericalsetup` function. The problem arises when transferring previous time-step data from the ode solver to the numerical setup function while solving a linear problem.  

Solution - 

The solution we arrived at was modifying the  `LinearStageOperator` struct to include `ws` and `usx`. The struct definition for the `NonlinearStageOperator` contains `usx` and `ws`. 

**I will link the code snippets below -** 

Linear Stage Operator - 
https://github.com/gridap/Gridap.jl/blob/ea403aa32d03b48f8eec0e5023ebadd5d8f6b8e6/src/ODEs/StageOperators.jl#L88-L92

Non-Linear Stage Operator - 
https://github.com/gridap/Gridap.jl/blob/ea403aa32d03b48f8eec0e5023ebadd5d8f6b8e6/src/ODEs/StageOperators.jl#L29-L35

Can this change be introduced in Gridap? This will make it possible to perform strong coupling for **linear** transient multiphysics problems. If this solution is not feasible, I would be happy to discuss alternatives. 
 
 Thanks, 
Shreyas

shreyas02:
Any thoughts @AlexandreMagueresse? 

I understand that adding those data fields increases memory requirements, which are not always desirable. I was wondering if we could add an input to the constructor that initiates the struct in two different ways. Suppose we pass LightWeight as an argument, which could be set as default; then the struct could initialise usx and ws as empty arrays. For specific use-cases when time integration data would be needed, all the data can be provided. 

I am throwing ideas around here to get a solution, and I look forward to hearing your thoughts.

PR #1089, opened 2025-03-21
Making ```LinearStageOperator``` struct  similar to ```NonlinearStageOperator``` struct

shreyas02:
The addition of this function allows users to overload this function so that they can modify the matrix with data from the ODE solver in the system solver. This is a solution I have come up with after discussing with @oriolcg. This solution is for issue #1075.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added the function `update_sysslvr!` in Gridap.ODEs. This gives the user the ability to modify the system solver by overloading this function. Currently implemented for Linear cases of GeneralizedAlpha1, GeneralizedAlpha2 and ThetaMethod.
+- Added the function `update_sysslvr!` in Gridap.ODEs. This gives the user the ability to modify the system solver by overloading this function. Currently implemented for Linear cases of GeneralizedAlpha1, GeneralizedAlpha2 and ThetaMethod. Since PR[#1089](https://github.com/gridap/Gridap.jl/pull/1089)
 
 ### Added
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added the function `update_sysslr!` in Gridap.ODEs. This gives the user the ability to modify the system solver by overloading this function. Currently implemented for Linear cases of GeneralizedAlpha1, GeneralizedAlpha2 and ThetaMethod.
+
+### Added
+
 - Added definition for `DiracDelta` with `Triangulation` argument. Since
   PR[#1087](https://github.com/gridap/Gridap.jl/pull/1087).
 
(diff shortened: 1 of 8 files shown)


JordiManyer:
I am sorry @shreyas02 , but this is very hacky. I will not merge this. 

I believe the proper way would be to have this function passed by the user as an input to the constructor, and to have `(args...) -> nothing` as default. But modifying the behaviour of something by having the user overwrite an internal function is just terrible code design, in my opinion.

shreyas02:
Thanks for the feedback @JordiManyer. I agree that this is hacky. 

You mean, I pass the function to the constructor and then add it to the ODE Solver struct, and then I call the function when required, keeping the default function to return nothing. 
 
 Even in this case, you would have to modify the struct definition, internal functions (where you call the function) and have a default function that returns nothing. This includes additional modifications to the struct. I wanted to change the code base as little as possible. If this sounds alright, I will add this. 
 
I don't think any of these solutions here look clean to me. However, this functionality is essential when implementing this type of coupling through Gridap. 

The other option is to modify the LinearStageOperator struct. You could pass an option through the ODESolver constructor and then add this data to the LinearStageOperator struct, as I initially mentioned in issue #1075.

JordiManyer:
One thing I don't fully understand is what this update to the solver can do, considering most structures (if not all) are immutable. Can I see an example of `update_sysslvr!` implementation?

shreyas02:
I currently have a field in the solver struct that is a dictionary. So, I add ```usx``` and ```ws``` as a key-value pair in the dict. Looking back, this function might be too specific for this use case.

Another good alternative is to update the sysslvrcache to include this data. Now, I feel that might be more generic and better.  What do you think? 

Looking forward to your feedback @JordiManyer.

JordiManyer:
If you are modifying your own solver, you don't need any of this: Since we return an iterator, you can just modify the parameters of your solver in between iterations, right? 

Take the typical usage for instance: 

```julia
    uh = solve(odesolver, ...)

    for (t, uh_t) in uh
       do stuff...
    end
```

you can simply do 

```julia
    uh = solve(odesolver, ...)

    for (t, uh_t) in uh
       update_solver!(my_solver)
    end
```

and it should modify your parameters just fine. No need to hack into the ODE solvers. 

Does this work for you?

shreyas02:
So I need ```ws``` and ```usx``` where ```usx``` changes every timestep. This data is only in the local scope in functions ```ode_march``` and ```ode_start```. I am looking for a way to access this data.

JordiManyer:
Then I guess what makes more sense is to also keep `tx`, `usx` and `ws` within the `LinearStageOperator` (like the `NonlinearStageOperator` does. It makes things more consistent accross the code, and would require minimal changes. 

So basically: 
- Add the fields to the struct
- Modify the (only) constructor that exists for it

Would this be acceptable?

shreyas02:
Yes, as you mentioned, this would be acceptable to me and require minimal code changes.  

I assumed this would increase the memory requirements, thus making it unnecessary for every transient problem.  If this is not a big issue, I can add this change.

JordiManyer:
The only things that could take up memory are the `usx`, but if you have a look at most solvers (GeneralizedAlpha1/2, RK, etc...) these are actually kept somewhere within the `odeslvrcache` or similar. Which means the memory is already being used, we are just adding pointers to it. So it should be fine.

shreyas02:
That sounds great. I'll add this here, then tomorrow morning. Thanks @JordiManyer!

JordiManyer:
@shreyas02 Thank you for the contribution!

Issue #927, opened 2023-08-04
Nedelec element on boundaries


HelgeGehring:
I'm trying to use a Nédélec element on a boundary, but the creation of the FESpace fails.
Is this an error in Gridap? Or am I just creating it wrong?

Thanks a lot for your help and the nice library!

Here a minimal example producing the error

```
using Gridap
using Gridap.Geometry

domain = (0, 2, 0, 2, 0, 2)
partition = (10, 10, 10)
oldmodel = CartesianDiscreteModel(domain, partition)

labels = get_face_labeling(oldmodel)
face_mask = get_face_mask(labels, ["boundary"], 2)
model = DiscreteModelPortion(DiscreteModel(Polytope{2}, oldmodel), face_mask)

reffe = ReferenceFE(lagrangian,Float64,1)
V = FESpace(model,reffe) # Works

reffe = ReferenceFE(nedelec,1)
V = FESpace(model,reffe) # Doesn't work
```

HelgeGehring:
ups, there was a small error in the example, I've updated it.

simbilod:
+1 this would be very useful!

HelgeGehring:
Here a simplified version which generates the same error:

I generated two meshes using (python code)
```
import gmsh

for z in [0,1]:
    gmsh.initialize()

    gmsh.model.add("t1")

    lc = 5
    gmsh.model.geo.addPoint(0, 0, z, lc, 1)
    gmsh.model.geo.addPoint(1, 0, z, lc, 2)
    gmsh.model.geo.addPoint(1, 1, z, lc, 3)
    gmsh.model.geo.addPoint(0, 1, z, lc, 4)

    gmsh.model.geo.addLine(1, 2, 1)
    gmsh.model.geo.addLine(3, 2, 2)
    gmsh.model.geo.addLine(3, 4, 3)
    gmsh.model.geo.addLine(4, 1, 4)

    gmsh.model.geo.addCurveLoop([4, 1, -2, 3], 1)
    gmsh.model.geo.addPlaneSurface([1], 1)

    gmsh.model.geo.synchronize()
    gmsh.model.addPhysicalGroup(1, [1, 2, 4], 5)
    gmsh.model.addPhysicalGroup(2, [1], name = "My surface")

    gmsh.model.mesh.generate(2)
    gmsh.write(f"t1-{z}.msh")

    gmsh.finalize()
```

[...]

PR #1094, opened 2025-04-08
Fix bug for issue #927

eazytbl:
Hi ! Now I need to use Nédélec elements on the boundary, just like in Issues #927 . 
I just need this step to complete my subsequent work, and I don't know much about manifolds. 
I just followed the steps in #577  to fix the bug. Perhaps my code also has some issues, but the results show that it is still okay.
Thanks a lot!

### DIFF
--- a/src/Polynomials/QGradMonomialBases.jl
+++ b/src/Polynomials/QGradMonomialBases.jl
@@ -21,7 +21,7 @@ Base.size(a::QGradMonomialBasis) = (_ndofs_qgrad(a),)
 # @santiagobadia : Not sure we want to create the monomial machinery
 Base.getindex(a::QGradMonomialBasis,i::Integer) = Monomial()
 Base.IndexStyle(::QGradMonomialBasis) = IndexLinear()
-return_type(::QGradMonomialBasis{D,T}) where {D,T} = T
+return_type(::QGradMonomialBasis{D,T}) where {D,T} = VectorValue{D,T}
 
 """
     QGradMonomialBasis{D}(::Type{T},order::Int) where {D,T}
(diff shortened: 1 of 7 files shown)


eazytbl (commented):
Change the order of the product, and Hcurltest passes

src/Polynomials/QGradMonomialBases.jl
```diff
@@ -21,6 +21,7 @@ Base.size(a::QGradMonomialBasis) = (_ndofs_qgrad(a),)
 # @santiagobadia : Not sure we want to create the monomial machinery
 Base.getindex(a::QGradMonomialBasis,i::Integer) = Monomial()
 Base.IndexStyle(::QGradMonomialBasis) = IndexLinear()
+return_type(::QGradMonomialBasis{D,T}) where {D,T} = T
```
Antoinemarteau, line 24: return_type here should return VectorValue{D,T} and VectorValue{D,Float64} for QGradMonomialBasis and NedelecPrebasisOnSimplex respectively.
I know this is inconsistant with return_type(::QCurlGradMonomialBasis), this one should also be VectorValue{D,T}, ideally you can fix those 3 in this PR.

JordiManyer: Checks are passing, so we should also add some tests for this ideally.

eazytbl: Thanks for the advice and help, I've now added some tests to the corresponding test file.

eazytbl: Thanks for the advice and help, I've now changed the return value type of return_type and added some tests to the corresponding test file.

PR #1114, opened 2025-06-19
Proposition: small Pushforward API rework for generality and flexibility

Antoinemarteau:
I feel like the current `PushforwardRefFE` type is not necessary, not mathematically accurate and may be limitting. Indeed, the geomap / trivial Piola map of functions is mathematically a pushforward, so any reffe is theoretically a pushforward-reffe.

These changes remove it by dispatching on `Pushforward(ReferenceFEName)` instead of on `ReferenceFEName` in Pullbacks.jl. I defined the trivial `IdentityPiolaMap` for completeness but it is skipped in `get_cell_dof_basis/shapefun`.

I also removed the conformity from arguments of `get_cell_pushforward`:
Indeed, if I understand correctly, the Piola map isn't necessary deducible from the conformity, because one could use DG method = `:L2` conformity to approximate H1 / Hcurl / Hdiv functions that use iddentity / Covariant / contravariant Piola, not the L2 Piola. 

So if we want to use e.g. the L2 Piola map (that only divides by det(J)) for our Lagrangian/Serandipity/Bezier elements, we will need `Pushforward` to depend on the `reffe` value rather than just its `ReferenceFEName` and conformity, and to dispatch on the `pushfoward` rather than on the `ReferenceFEName`.

What do you think @JordiManyer ?


JordiManyer:
So a couple comments: 
- I would not create the IdentityPiolaMap if it is not needed. 
- In the case of L2 conformity, I am not sure the piola maps are needed. That's why I had that there before. I believe its the same behaviour we had in the original code. Maybe @santiagobadia can comment.

Antoinemarteau:
Ok for Identity Piola. So, when using discontinuous galerkin for curl/div conforming fields, Dirichlet boundary/periodicity conditions aren't set strongly either ?  Because to do this, you need to properly map the shape functions owned by the boundary elements, no ?

JordiManyer:
If you are in l2, dirichlet bcs dont make sense. We talked with santi today and agreed we should leave it as it was.

Antoinemarteau:
Okey, I didn't understand properly then, thanks.

So if I revert everything except geting rid of `PushforwardRefFE`, would you merge this, or do I just close this issue ?

JordiManyer:
Another comment is that we should not be defining `Pushforward(name)` for each reffe type. We should instead have the identity as a default, directly on the `ReferenceFEName` type. Then users can override the default behaviour.

Antoinemarteau:
I reverted unwanted changes as discussed.

But actually the `conformity` argument is ignored in `get_cell_pushforward` methods for `[Co/Contra]VariantPiolaMap`, so the Piola map are currently applied even for L2Conformity, the behavior isn't the same as master. (Actually `conformity` is completely ignored in FESpaces/Pullbacks.jl).

It could be fixe as follow:
```
Pushforward(reffename, ::Conformity) = Pushforward(reffe)
Pushforward(reffename, ::L2Conformity) = nothing  # or IddentityPiolaMap if I define it.
```
and then `conformity`  is not needed/passed into `get_cell_pushforward`.

On master, the conformity is part of`typeof(reffe)`, but not on moment-based-reffes. In find it cleaner to get rid of trivial Piola map before going into `get_cell_pushforward` in general.

JordiManyer:
> In find it cleaner to get rid of trivial Piola map before going into get_cell_pushforward in general.
I agree, I don't like the `isnothing` checks within the routines. We can keep the IdentityPiolaMap.

Antoinemarteau:
EDIT: I added `IdentityPiolaMap` and the following to fix L2 conformity
```julia 
Pushforward(reffename, ::Conformity) = Pushforward(reffe)
Pushforward(reffename, ::L2Conformity) =IdentityPiolaMap()
```
I still feel that it could be better to implement `Pushforward` on the `::ReferenceFE` instances, rather than their static`ReferenceFEName` and conf. only, especially if we want to easily implement several scaling tricks similar to #1120 .
The `::Pushforward` of a reffe could be stored in `metadata` field of `GenericReferenceFE`, or a new field...

JordiManyer:
I think the `metadata` is already being used? In any case, if this is something all reffes are gonna have, I would probably add it as a new field. It would be nice, however, to preserve the current constructors. So we could add it as an arg with default value or something.

JordiManyer:
@Antoinemarteau is this mergeable? Or do you want to have a look at the last comments?

Antoinemarteau:
Yeah lets merge it as is, and do the other change if they happen to be necessary later...

PR #1118, opened 2025-06-20
Update norm

### LINKED ISSUE #1116: Norm of a VectorValue
Hi,

I think that the norm of a complex VectorValue is not well defined:

`norm(VectorValue(1.0,1.0,im))` returns `1.0 + 0.0im` but  `LinearAlgebra.norm([1.0,1.0,im])` returns `1.7320508075688772`

This is because` norm(v)` is defined as `sqrt(inner(v,v))` but it should be defined as `sqrt(real(inner(v,conj(v))))` to obtain the same result as `LinearAlgebra.jl`.

Is there a reason why the norm is defined like that?

ovanvincq:
Update norm function to be compatible with Complex Vector or Tensor. 
Fix #1116 

Before: `norm(VectorValue(1.0,1.0,im))` returns `1.0 + 0.0im`
After: `norm(VectorValue(1.0,1.0,im))` returns `1.7320508075688772`

@JordiManyer : normally, I've specialized the function so that there's no impact on real vectors/tensors.

### DIFF
--- a/test/TensorValuesTests/OperationsTests.jl
+++ b/test/TensorValuesTests/OperationsTests.jl
@@ -742,8 +742,13 @@ v = VectorValue(2.0,3.0)
 @test dot(u,v) ≈ inner(u,v)
 @test norm(u) ≈ sqrt(inner(u,u))
 
+u = VectorValue(1.0,im)
+@test norm(u) ≈ sqrt(2)
+
 a = TensorValue(1,2,3,4)
+a2 = TensorValue(1,im,1,im)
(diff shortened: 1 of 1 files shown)


JordiManyer:
@ovanvincq Thank you for the contribution!  Looks good to me! Could you add some tests and add an entry to NEWS.md?

ovanvincq:
@JordiManyer done!

JordiManyer:
Thank you again for the contribution @ovanvincq !

PR #1120, opened 2025-06-27
Scaled contravariant piola map for Div-conforming elements

amboschman:
This PR allows the user to specify whether a HDiv-conforming element (e.g. Raviart-Thomas or BDM) should be used with the (default) contravariant Piola map or a rescaled version through the use of an extra (optional) argument in `FESpace` when constructing an HDiv-conforming finite element space, e.g. `FESpace(model, reffe_RT, conformity=:HDiv, contra_variant_piola_map_type=ScaledContraVariantPiolaMap())`. The scaled contravariant Piola map is defined as the standard contravariant Piola map for which the determinant of the Jacobi matrix is rescaled using `^{1/D}` (with `D` the `D1`-dimensional volume of the parallelepiped formed by the rows of `J`). The use of this scaled contravariant Piola map can result in better conditioning of certain (mixed) problems, such as the Darcy equation.

For this to work,
- the interface of `CellFE` has been updated to not only pass along `args` but also `kwargs` to `get_cell_shapefuns` and `get_cell_dof_basis`, so that the user can (optionally) specify the `contra_variant_piola_map_type` in `FESpace`;
- the `scaled_meas` operation has been introduced to define the scaled version of the contravariant Piola map;
- the `lazy_map`'s and `evaluate!`'s are changed in `RaviartThomasRefFEs.jl`so that either the default operation `ContraVariantPiolaMap()` or the selected (`ContraVariantPiolaMap()` or `ScaledContraVariantPiolaMap()`) is performed.

### DIFF
--- a/src/FESpaces/DivConformingFESpaces.jl
+++ b/src/FESpaces/DivConformingFESpaces.jl
@@ -29,7 +29,8 @@ struct TransformRTDofBasis{Dc,Dp} <: Map end ;
 function get_cell_dof_basis(model::DiscreteModel,
                             cell_reffe::AbstractArray{<:GenericRefFE{<:DivConforming}},
                             ::DivConformity,
-                            sign_flip=get_sign_flip(model, cell_reffe))
+                            map_type=nothing)
+    sign_flip = get_sign_flip(model, cell_reffe)
     cell_map  = get_cell_map(Triangulation(model))
     phi       = cell_map[1]
     Jt        = lazy_map(Broadcasting(∇),cell_map)
@@ -48,9 +49,16 @@ end
 function get_cell_shapefuns(model::DiscreteModel,
                             cell_reffe::AbstractArray{<:GenericRefFE{<:DivConforming}},
                             ::DivConformity,
-                            sign_flip=get_sign_flip(model, cell_reffe))
+                            map_type=nothing)
+    sign_flip=get_sign_flip(model, cell_reffe)
     cell_reffe_shapefuns=lazy_map(get_shapefuns,cell_reffe)
-    k=ContraVariantPiolaMap()
+
+    if map_type == nothing
+      k=ContraVariantPiolaMap()
+    else 
+      k=map_type
+    end
+    
     lazy_map(k,
              cell_reffe_shapefuns,
              get_cell_map(Triangulation(model)),
--- a/src/FESpaces/DivConformingFESpaces.jl
+++ b/src/FESpaces/DivConformingFESpaces.jl
@@ -29,8 +29,9 @@ struct TransformRTDofBasis{Dc,Dp} <: Map end ;
 function get_cell_dof_basis(model::DiscreteModel,
                             cell_reffe::AbstractArray{<:GenericRefFE{<:DivConforming}},
                             ::DivConformity,
-                            map_type=nothing)
-    sign_flip = get_sign_flip(model, cell_reffe)
+                            args...; 
+                            sign_flip=get_sign_flip(model, cell_reffe),
+                            kwargs...)
     cell_map  = get_cell_map(Triangulation(model))
     phi       = cell_map[1]
     Jt        = lazy_map(Broadcasting(∇),cell_map)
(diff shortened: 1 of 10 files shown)


src/FESpaces/DivConformingFESpaces.jl
```diff
@@ -48,10 +50,13 @@ end
 function get_cell_shapefuns(model::DiscreteModel,
                             cell_reffe::AbstractArray{<:GenericRefFE{<:DivConforming}},
                             ::DivConformity,
-                            sign_flip=get_sign_flip(model, cell_reffe))
+                            args...; 
+                            sign_flip=get_sign_flip(model, cell_reffe),
+                            contra_variant_piola_map_type=ContraVariantPiolaMap(),
```
amartinhuertas, line 55: For consistency, annotate the kw-arg with `::ContraVariantPiolaMapType` as you did in the other source file.

amartinhuertas:
Hi @amboschman  ... thanks for your work! It looks good. 

The only major thing I am missing is a test that stresses the `ScaledContraVariantPiolaMap` (i.e., the new kw-arg added to the FESpace). 

Perhaps you may some tests here: 
https://github.com/gridap/Gridap.jl/blob/master/test/FESpacesTests/DivConformingFESpacesTests.jl

Also, have we tried this with (1) BDM. (2) RT elements on triangles?

src/FESpaces/DivConformingFESpaces.jl
amboschman: Thanks for flagging this!

JordiManyer:
Why did we do this in master? I would be working directly on the new moment based branch, otherwise we are doing redundant work. Also, the implementation in the new branch could be done much more elegantly by overwriting the pushforward instead of propagating a kwarg all the way into the depths of the code.

amboschman:
Thanks Jordi for looking at this.

> Why did we do this in master? I would be working directly on the new moment based branch, otherwise we are doing redundant work. 

I had a very brief chat with Antoine, but he mentioned that one or two things still had to be finished, so it might take ~ 1 month(? if I remember correctly) for it to be released. Ideally, I would like to use this switch within a shorter timeframe. But happy to learn what can be done to better coordinate this, or contribute/develop a similar thing in the moment-based branch.


> Also, the implementation in the new branch could be done much more elegantly by overwriting the pushforward instead of propagating a kwarg all the way into the depths of the code.

That would not give the option to the user to switch between the two different ContraVariantPiolaMappings. I wanted the standard one to be the default, as that is what a user would expect this code to do.

amboschman:
> The only major thing I am missing is a test that stresses the ScaledContraVariantPiolaMap (i.e., the new kw-arg added to the FESpace).

I fully agree. 

> Perhaps you may some tests here:
https://github.com/gridap/Gridap.jl/blob/master/test/FESpacesTests/DivConformingFESpacesTests.jl

Looking at this at the moment. At a quick glance, having a good test with  `interpolate` and `DIV` in combination with the `ScaledContraVariantPiolaMap` require some more work...

> Also, have we tried this with (1) BDM. (2) RT elements on triangles?

I tested this once quickly on my system (h-convergence test), but will include this in a test.

amartinhuertas:
> Why did we do this in master? 
1. Because it is a relatively simple change.
1. Because we need it relatively urgently for a code base which currently relies on `Gridap#master`.
2. Because (we thought) it could also be useful for other users via a non-breaking patch release while they familiarize themselves with the new machinery for moment-based RefFEs.

> I would be working directly on the new moment based branch, otherwise we are doing redundant work. Also, the implementation in the new branch could be done much more elegantly by overwriting the pushforward instead of propagating a kwarg all the way into the depths of the code.

Anyway, as you say, we also need to implement this in that branch. It would be great if you can assist us with additional details on how to do this. Thanks!

JordiManyer:
> That would not give the option to the user to switch between the two different ContraVariantPiolaMappings. I wanted the standard one to be the default, as that is what a user would expect this code to do.

I wanted to keep both as well, just override the pushforward on the kwarg, directly on the reffe instead of propagating it to the functions. 

> I had a very brief chat with Antoine, but he mentioned that one or two things still had to be finished, so it might take ~ 1 month(? if I remember correctly) for it to be released. Ideally, I would like to use this switch within a shorter timeframe. But happy to learn what can be done to better coordinate this, or contribute/develop a similar thing in the moment-based branch.

Sure, but you can always use the branch. It doesn't need to be released for you to be able to use it in another package.

JordiManyer:
Anyway, lets just add some tests and merge this as it is, with the understanding it will get modified soon.

Antoinemarteau:
Who are those potential users? 

The user API implemented here could change in next minor if:
- this scaling is generalized to Curl-conforming elements / all piola maps, so renaming the kwarg will be needed
- the choice of Piola map is attached to the reffe rather than fespace, like currently done on `moment-based-reffes`

So I'm not sure if its a good idea to let users adopt a possibly temporary API design. But if its mostly for you of course this is not an issue..

amartinhuertas:
Let us put this PR on hold ... There are still several issues that we do not understand and thus not ready to be merged. 

I have transformed the PR into a draft PR.

Antoinemarteau:
I suggest closing this PR, the re-scaling of Piola-mapped elements has been implemented by the `scale_dof` and `global_meshsize` kwargs for `FESpace(...)` introduced in #1048.

amartinhuertas:
yes, I agree, this should be closed.

PR #1121, opened 2025-06-27
GridapMakie as an extension to Gridap

ovanvincq:
Hi,

This PR is a proposal to transform `GridapMakie.jl` into an extension of `Gridap.jl`.
I named this extension `GridapMakieExt`. 

It is compatible with `Makie.jl` v0.24 and not with previous versions because the recipes use the new `ComputeGraph` interface.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Gridap"
 uuid = "56d4f2e9-7ea1-5844-9cf6-b9c51ca7ce8e"
 authors = ["Santiago Badia <santiago.badia@monash.edu>", "Francesc Verdugo <f.verdugo.rojano@vu.nl>", "Alberto F. Martin <alberto.f.martin@anu.edu.au>"]
-version = "0.20.2"
+version = "0.20.3"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"


ovanvincq:
Tests fail because:
- Makie requires at least julia v1.10
- ci.yml should be modified to activate OpenGL capabilities

JordiManyer:
Hi @ovanvincq thank you for the work. 

I do have some preliminary comments: 
- I think it is extremely important we acknowledge the original creators of the package. I would add a clear admonition block at the top of the documentation page for the extension referencing the original package, the original authors and the fact that is was created during a Google SummerOfCode. 
- I would put the images into a subfolder of assets. 
- In terms of organizing the documentation, I would add a separator for extensions, and list the file as `Makie extension` instead. There will be some major documentation overhaul soon-ish, so it will probably be moved around again, but I think that should be ok for now. 
- Removing compatibility for 1.9 is OK, I think. 
- I see what you mean by the OpenGL stuff. In light of this, I would split the tests into their own CI job, and just copy-paste the CI file we have in the package. I think I have some ideas on how to unify all of this, but splitting like this should be ok for now.

ovanvincq:
Hi @JordiManyer 

I made the changes in the documentation.

For OpenGL, I think you need to change `run: julia...` to `run: DISPLAY=:0 xvfb-run -s '-screen 0 1024x768x24' julia...` in `ci.yml`
If you prefer, I can also include `CairoMakie` in the weak dependencies and run the test with `CairoMakie`

A quick question: does a Triangulation{0,N} with N>1 exist?

JordiManyer:
As I said, we can split the tests and run the ones we already had in GridapMakie.jl, I'll unify them again in a later PR. Basically copy paste the [GridapMakie CI](https://github.com/gridap/GridapMakie.jl/blob/ebb748f68c92eda351d3c326bdb5db5225f9338f/.github/workflows/ci.yml#L25). You could also do the CairoMakie workaround, of it works. 

> A quick question: does a Triangulation{0,N} with N>1 exist?

Yes, a triangulatin of points would be Triangulation{0,D} with D the ambient dimension. I don't know if anyone would ever want to use it, but that is what it would be. I would say it's not super important to support that on Makie.

ovanvincq:
@JordiManyer I copied the CI file from GridapMakie and the test succeeded

PR #1126, opened 2025-07-06
Enable non-surjective n2o maps for AdaptivityGlue

amartinhuertas:
Fix AdaptivityGlue such that it is now able to work with non-surjective n2o maps as per-required by GridapP4est whenever we have more than one ghost layer


amartinhuertas:
Related to https://github.com/gridap/GridapP4est.jl/pull/74

Issue #1129, opened 2025-07-09
Sum of MultiField bases requires additional API


zjwegert:
Hi @dhtantoy and @JordiManyer,
The new `MultiFieldFEBasesAddition` breaks for the following case
```julia
using Gridap
using Gridap.MultiField

model = CartesianDiscreteModel((0,1,0,1),(4,4))
Ω = Triangulation(model)
dΩ = Measure(Ω,2)
V = FESpace(Ω,ReferenceFE(lagrangian,Float64,1))
VV = MultiFieldFESpace([V,V])
du1u2 = get_fe_basis(VV)
F((du1,du2)) = ∫(du1 + du2)dΩ
F(du1u2)
```
and produces this error
```julia
ERROR: MethodError: no method matching CellField(::Gridap.MultiField.MultiFieldFEBasesAddition{…}, ::Gridap.Geometry.BodyFittedTriangulation{…}, ::ReferenceDomain)
The type `CellField` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  CellField(::Any, ::Triangulation)
   @ Gridap C:\Users\Zac\.julia\dev\Gridap\src\CellData\CellFields.jl:114
  CellField(::CellField, ::Triangulation, ::DomainStyle)
   @ Gridap C:\Users\Zac\.julia\dev\Gridap\src\CellData\CellFields.jl:110
  CellField(::Number, ::Triangulation, ::DomainStyle)
   @ Gridap C:\Users\Zac\.julia\dev\Gridap\src\CellData\CellFields.jl:94
  ...

Stacktrace:
 [1] integrate(a::Gridap.MultiField.MultiFieldFEBasesAddition{…}, quad::CellQuadrature{…})
   @ Gridap.CellData C:\Users\Zac\.julia\dev\Gridap\src\CellData\CellQuadratures.jl:173
 [2] integrate(f::Gridap.MultiField.MultiFieldFEBasesAddition{…}, b::Gridap.CellData.GenericMeasure)
   @ Gridap.CellData C:\Users\Zac\.julia\dev\Gridap\src\CellData\DomainContributions.jl:164
 [3] *(a::Gridap.CellData.Integrand, b::Gridap.CellData.GenericMeasure)
   @ Gridap.CellData C:\Users\Zac\.julia\dev\Gridap\src\CellData\DomainContributions.jl:148
 [4] F(::MultiFieldCellField{ReferenceDomain})
[...]

PR #1124, opened 2025-07-03
Add bubble element support for TRI and TET

dhtantoy:
### Summary
This PR adds support for bubble elements on both TRI and TET reference cells. It enables solving the Stokes equations using the MINI element, as discussed in [1123](https://github.com/gridap/Gridap.jl/issues/1123).

### Convergence Test
I have tested the convergence order of the implementation. The test code are provided below:
```julia
function main(N)
	u((x, y)) = VectorValue(
		sin(π*x)^2*sin(2π*y),
		-sin(π*y)^2*sin(2π*x),
	)
	p((x, y)) = cos(π*x)*cos(π*y)
	f((x, y)) = VectorValue(
		-cos(π*y)*sin(π*x)*π - 2*cos(π*x)^2*sin(2π*y)*(π^2) + 6*sin(2π*y)*(sin(π*x)^2)*(π^2),
		-cos(π*x)*sin(π*y)*π + 2*cos(π*y)^2*sin(2π*x)*(π^2) - 6*sin(2π*x)*(sin(π*y)^2)*(π^2),
	)

	m = simplexify(CartesianDiscreteModel((0, 1, 0, 1), (N, N)))
	trian = Triangulation(m)
	dx = Measure(trian, 3)

	B_test = TestFESpace(m, ReferenceFE(bubble, VectorValue{2, Float64}))
	B_trial = TrialFESpace(B_test)
	U_test = TestFESpace(m, ReferenceFE(lagrangian, VectorValue{2, Float64}, 1); conformity = :H1, dirichlet_tags = "boundary")
	U_trial = TrialFESpace(U_test, u)
	P_test = TestFESpace(m, ReferenceFE(lagrangian, Float64, 1); conformity = :H1, dirichlet_tags = "boundary")
	P_trial = TrialFESpace(P_test, p)
	X = MultiFieldFESpace([U_trial, B_trial, P_test])
	Y = MultiFieldFESpace([U_test, B_test, P_trial])

	a((u, p), (v, q)) = ∫(∇(v) ⊙ ∇(u) - (∇ ⋅ v) * p + q * (∇ ⋅ u))dx
	l((v, q)) = ∫(v ⋅ f)dx

	b_a((uc, ub, p), (vc, vb, q)) = a((uc + ub, p), (vc + vb, q))
	b_l((vc, vb, q)) = l((vc + vb, q))

	op = AffineFEOperator(b_a, b_l, X, Y)
	uch, ubh, _ = solve(LUSolver(), op)
[...]

PR #1130, opened 2025-07-13
Fix a bug in the summation logic of  `MultiFieldFEBasisComponent`

DimhamT:
## Summary
Hi. This PR fix the bug discussed in the [comment](https://github.com/gridap/Gridap.jl/pull/1124#issuecomment-3036512733) and also related to the issue #1129. It removes the `MultiFieldFEBasesAddition` type introduced in #1124 and implements some internal interfaces directly for `ArrayBlock`.

## Implementation Details
1. The summation logic in `MultiFieldFEBasisComponent` relies on the function `evaluate!(cache,::Broadcasting{<:Operation},::ArrayBlock,::ArrayBlock)`. This has now been redirected to `evaluate!(cache, ::BroadcastingFieldOpMap, ::ArrayBlock, ::ArrayBlock)`, which already provides a reusable implementation internally.
2. The `evaluate!` function relies on the presence of the `testvalue(:Type{<:Field})` interface [here](https://github.com/gridap/Gridap.jl/blob/0b2e2a7b3422e7a5b4ffb9f3d01cc76171475c1e/src/Fields/FieldArrayBlocks.jl#L433-L434). To support this, definitions of `testvalue` have been added for several relevant subtypes of `Field`.
3. `zero(::Type{<:Field})` has been defined to support evalution involving `ZeroBlockMap` [here](https://github.com/gridap/Gridap.jl/blob/0b2e2a7b3422e7a5b4ffb9f3d01cc76171475c1e/src/Arrays/ArrayBlocks.jl#L377)

### DIFF
--- a/src/Fields/FieldArrays.jl
+++ b/src/Fields/FieldArrays.jl
@@ -518,8 +518,8 @@ function testitem(a::BroadcastOpFieldArray)
   return_value(Operation(a.op),fs...)
 end
 for op in (tr,dot,:-)
-  @eval function testvalue(::Type{BroadcastOpFieldArray{typeof($op),T,N,A}}) where {T,N,A}
-    fs = tuple((testvalue(fi) for fi in A.parameters)...)
+  @eval function testvalue(::Type{BroadcastOpFieldArray{typeof($op),T,N,A}}) where {T,N,A<:Tuple}
+    fs = map(testvalue,fieldtypes(A))
     BroadcastOpFieldArray($op,fs...)
   end
 end
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.19.xx]
+
+### Fixed
+
+- `MultiFieldFEBasisComponent` now supports addition and subtraction. Since PR[#1130](https://github.com/gridap/Gridap.jl/pull/1130)
+
 ## [0.19.2] - 2025-07-08
 
 ### Added
--- a/src/Fields/FieldsInterfaces.jl
+++ b/src/Fields/FieldsInterfaces.jl
@@ -313,8 +313,10 @@ struct OperationField{O,F} <: Field
   fields::F
 end
 
-function testvalue(::Type{OperationField{O,F}}) where {O,F}
-  OperationField(testvalue(O),tuple((testvalue(f) for f in F.parameters)...))
+function testvalue(::Type{OperationField{O,F}}) where {O<:Field,F<:Tuple}
+  op = testvalue(O)
+  fields = map(testvalue,fieldtypes(F))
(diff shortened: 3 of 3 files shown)


JordiManyer:
Hi @dhtantoy , thank you for this. 

I now see I had actually already encountered (and fixed) some of this stuff. I had some fixes scattered in other repos. I think in general the fixes look good. The only comment would be that I think my fix for `testvalue` on `OperationField` is slightly more elegant: 

```
function Arrays.testvalue(::Type{Fields.OperationField{O,F}}) where {O<:Field,F<:Tuple}
  op = testvalue(O)
  fields = map(testvalue,fieldtypes(F))
  Fields.OperationField(op,fields)
end
```

I basically make use of the Julia API to get the types of the elements of a tuple. You should try to rewrite your fixes using that, if possible.

With the above rewrite and adding a NEWS entry I will gladly merge this.

DimhamT:
Hi, @JordiManyer . Thanks for the clarification! I'll update the `testvalue` definition to match you version, and add a NEWS entry as well.

src/Fields/FieldArrays.jl
```diff
@@ -536,6 +560,24 @@ return_value(a::BroadcastingFieldOpMap,args::AbstractArray...) = return_value(Br
 return_cache(a::BroadcastingFieldOpMap,args::AbstractArray...) = return_cache(Broadcasting(a.op),args...)
 evaluate!(cache,a::BroadcastingFieldOpMap,args::AbstractArray...) = evaluate!(cache,Broadcasting(a.op),args...)
 
+# The following code is implemented to store `Transpose`
+# only for `Field`, which is needed when using `CachedArray`.
+function testitem(
```
JordiManyer, line 565: Could you briefly comment on why this is necessary, and what the added function does?

dhtantoy: Hi, @JordiManyer . Sorry for not explaining this more clearly earlier.
In the default `testitem` implementation,it may return `first(a)`. However, the implementation `return_cache(::BroadcastingFieldOpMap,::AbstractArray{<:Field},::AbstractArray{<:Field})` in this PR would lose the `Transpose` type information, which may lead to a type assertion failure in some contexts. We could also implement `return_cache` and `evaluate!` for `BroadcastingFieldOpMap` and `ZeroBlockMap` to preserve the `Tranpose` type. The approach in this PR is relatively direct, and an even simple alternative would be to modify the default `testitem` as shown below.
```julia
function testitem(a::LazyArray{A,T} where A) where T
  if length(a) > 0 && isa(first(a),T)
    first(a)
  else
    gi = testitem(a.maps)
    fi = map(testitem,a.args)
    return_value(gi,fi...)
  end::T
end
```
But if you think this change could have unintended consequences, I'm happy to instead just add a comment to the new `testitem` above to explain the reasoning. Let me know if you have any further feedback!

JordiManyer: Ok this is quite serious then. The real problem is that `T` should always be `eltype(a)`, and therefore `first(a)` should always be of type `T`. Is this doesn't happen, there is something weird going on somewhere. 
Could you please open a new issue with a small reproducer where this case arose? I can try to have a look and find the root cause.

PR #1144, opened 2025-07-28
[moment-based-reffes] FEEC APIs, Bernstein bases for FEEC spaces on simplices, small Polynomials cleaning

Antoinemarteau:
Implemented FEEC polynomial bases of Arnold et al on simplices in any dimension, polynomial and form order. The P⁻ type is `BarycentricPmΛBasis` and the P type is `BarycentricPΛBasis`. Currently, only the vector proxy of the basis forms is exposed, implementing the exterior algebra operation would be needed to use the underlying form valued polynomial. The implementation is documented in detail in the "Bernstein bases algorithms" dev notes. In term of performance, these bases are on par with Monomial counterpart for basic evaluate, and generally faster for derivatives. The compilation for polynomial order >12-15 is very slow (minutes) though due to the @generated functions.

New high level `ReferenceFE`s constructor using Arnold et al FEEC notations (Periodic Table of the Finite Elements): `ReferenceFE(F::Symbol, r, k, [, T::Type]; kwargs...)` with `F ∈ {:P⁻,:P,:Q⁻,:S}` the element family, `r` polynomial order and `k` the form order.

New high level Polynomial basis factory `FEEC_poly_basis` that follows the FEEC notations. It replaces the former `[P/Q][Curl]GradBasis` factories. The default poly type is `Bernstein` on simplices and `Legendre` on n-cube.

Defined a geometric decomposition API, consisting in the methods `has_geometric_decomposition`, `get_face_own_funs` and `get_facet_flux_sign_flip`. It is implemented for `BernsteinBasisOnSimplex` and `BarycentricP(m)ΛBasis`.
Moved the `::Conformity` definitions in `ReferenceFEInterfaces.jl` because I need them in the geometric decomposition APIs.

Reorganized a bit the MomentBasedReferenceFEs.jl file to split the dof basis computation and the RefFE creation. The `sh_is_pb` keyword -- for "shape functions is pre-basis" -- is available in the `ReferenceFE` constructors to use the pre-basis directly as shape function, leading to a dof change of basis insead of a shapefun change of basis. This requires the basis to fullfill the geometric decomposition API.

On simplices, the default polynomial bases of BDM, Nédélec and Raviart-Thomas RefFEs have changed for `BarycentricP(m)ΛBasis` bases. They also use the `sh_is_pb=true` by default on simplices, leading to better conditioned systems matrices for higher order.

Similarly, on n-cubes, the default polynomial bases of BDM, Nédélec and Raviart-Thomas RefFEs have changed for `Legendre` (tensor-product)  bases instead of `Monomial`, I didn't compare conditioning though.

Minor `MultiValue` bugfixes for `isless` and `<=` with scalars and some `(Skew)SymTensorValue` operations.  Sorry, most of `Operation.jl` changes is an auto-typesetting change I inadvertently pushed.

Changed `Base.==` for `ReferenceFE`s, and implemented it for `LinearCombinationFieldVector` and `LinearCombinationDofVector`. The implementation now uses AutoHashEquals.jl, so RefFEs are now only equal if they have `==` prebasis, shapefun and dofs.


JordiManyer:
Have you pulled from moment-based reffes?

Antoinemarteau:
> Have you pulled from moment-based reffes?

Yes, and this PR goes into it.

PR #1152, opened 2025-08-18
[Moment based reffes] Fix tensor product bases and Polynomials optimization

Antoinemarteau:
There was a bug in `CompWiseTensorPolyBasis` (the basis for elements on n-cubes), it implicitely assumed the basis to be hierarchichal.
I want to lift this assumption because the geometrically decomposed 1D bases (ModalC0 and Bernstein) are not  (and cannot be) hierarchical between at least order 0 and 1.
 
Also fixed stuff I messed up when refactoring Polynomials and removing the polynomial order from the struct parameters, mainly the runtime dispatch in `evaluate!(c, ::PolynomialBasis, ...)` and the type of the fields of `CompWiseTensorPolyBasis`.


Antoinemarteau:
I cannot reproduce this "CI/Tests 1.10 Ubuntu x86" failure locally, with same julia version and test run command. TikzPictures fails to compile in the CI (not on my computer), might this messes the rest of the tests indirectly ?

PR #1158, opened 2025-09-09
Added isapprox support for MultiValue

miguelmaso:
This PR is a proposal to extend the interface of `MultiValue`. Its implementation follows the pattern of #1157.

**Current behaviour**
Right now, the function `isapprox(::MultiValue, ::MultiValue)` is available with default arguments, but is not possible to pass optional arguments like tolerance.

**Proposed behaviour**
After this PR, it will be able to call the function `isapprox` with optional arguments, e.g. `isapprox(A,B,rtol=eps(1.0)`.

### DIFF
--- a/src/TensorValues/Operations.jl
+++ b/src/TensorValues/Operations.jl
@@ -5,10 +5,10 @@
 (==)(a::MultiValue,b::MultiValue) = false
 (==)(a::MultiValue{S},b::MultiValue{S}) where {S} = a.data == b.data
 isapprox(a::MultiValue,b::MultiValue;kwargs...) = isapprox(get_array(a),get_array(b);kwargs...)
-(≈)(a::MultiValue{S,T1,N,0} where T1,b::MultiValue{S,T2,N,0} where T2;kwargs...) where {S,N} = true
+isapprox(a::MultiValue{S,T1,N,0} where T1,b::MultiValue{S,T2,N,0} where T2;kwargs...) where {S,N} = true
 
-function (≈)(
-  a::AbstractArray{<:MultiValue}, b::AbstractArray{<:MultiValue})
+function isapprox(
+  a::AbstractArray{<:MultiValue}, b::AbstractArray{<:MultiValue};kwargs...)
   if size(a) != size(b); return false; end
   for (ai,bi) in zip(a,b)
     if !(ai≈bi); return false; end
--- a/src/TensorValues/Operations.jl
+++ b/src/TensorValues/Operations.jl
@@ -6,6 +6,7 @@
 (==)(a::MultiValue{S},b::MultiValue{S}) where {S} = a.data == b.data
 (≈)(a::MultiValue{S},b::MultiValue{S}) where {S} = isapprox(get_array(a), get_array(b))
 (≈)(a::MultiValue{S,T1,N,0} where T1,b::MultiValue{S,T2,N,0} where T2) where {S,N} = true
+isapprox(a::MultiValue,b::MultiValue;kwargs...) = isapprox(get_array(a),get_array(b);kwargs...)
 
 function (≈)(
   a::AbstractArray{<:MultiValue}, b::AbstractArray{<:MultiValue})
@@ -31,8 +32,6 @@ end
 
 isless(a::Number,b::MultiValue) = all(isless.(a, b.data))
 isless(a::MultiValue,b::MultiValue) = @unreachable "Comparison is not defined between tensor of order greater than 1"
-
-isapprox(a::MultiValue,b::MultiValue;kwargs...) = isapprox(get_array(a),get_array(b);kwargs...)
 
 ###############################################################
 # Addition / subtraction
--- a/src/TensorValues/Operations.jl
+++ b/src/TensorValues/Operations.jl
@@ -5,6 +5,7 @@
 (==)(a::MultiValue,b::MultiValue) = false
 (==)(a::MultiValue{S},b::MultiValue{S}) where {S} = a.data == b.data
 isapprox(a::MultiValue,b::MultiValue;kwargs...) = isapprox(get_array(a),get_array(b);kwargs...)
+(≈)(a::MultiValue{S,T1,N,0} where T1,b::MultiValue{S,T2,N,0} where T2;kwargs...) where {S,N} = true
 
 function (≈)(
   a::AbstractArray{<:MultiValue}, b::AbstractArray{<:MultiValue})
--- a/src/TensorValues/Operations.jl
+++ b/src/TensorValues/Operations.jl
@@ -4,13 +4,13 @@
 
 (==)(a::MultiValue,b::MultiValue) = false
 (==)(a::MultiValue{S},b::MultiValue{S}) where {S} = a.data == b.data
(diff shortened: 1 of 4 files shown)


src/TensorValues/TensorValues.jl
```diff
@@ -78,7 +78,7 @@ export indep_components_names
 
 import Base: show
 import Base: zero, one
-import Base: +, -, *, /, \, ==, ≈, isless
+import Base: +, -, *, /, \, ==, ≈, isless, isapprox
```
Antoinemarteau, line 81: Not needed, `isapprox === ≈` is `true`.

src/TensorValues/Operations.jl
```diff
@@ -32,6 +32,8 @@ end
 isless(a::Number,b::MultiValue) = all(isless.(a, b.data))
 isless(a::MultiValue,b::MultiValue) = @unreachable "Comparison is not defined between tensor of order greater than 1"
 
+isapprox(a::MultiValue,b::MultiValue;kwargs...) = isapprox(get_array(a),get_array(b);kwargs...)
```
Antoinemarteau, line 35: I would put this method line 7 of this file (replace the previous method, that is redundant).

test/TensorValuesTests/OperationsTests.jl
```diff
@@ -31,6 +31,8 @@ b = VectorValue(1,3,3)
 @test iszero(VectorValue(1,2,3) - VectorValue(1.0,2.0,3.0))
 @test iszero(zero(VectorValue(1,2,3)))
 @test isapprox(VectorValue(1,2,3), VectorValue(1.0,2.0,3.0))
+@test isapprox(VectorValue(1,2,3), VectorValue(1.0,2.0,3.0), atol=eps(1.0))
+@test isapprox(VectorValue(1,2,3), VectorValue(1.0,2.0,3.0), rtol=eps(1.0))
```
Antoinemarteau, line 35: Would you mind adding a test for !isapprox on two tensors of different shape ?

Antoinemarteau (changes requested):
Hello Miguel, thanks for your PR :) 

I gave some comments, I think this can be merged after taking them into account and updating README.md.

miguelmaso:
> I gave some comments, I think this can be merged after taking them into account and updating README.md.

Added to NEWS.md

src/TensorValues/TensorValues.jl
miguelmaso: If removed, I get an error shouldimport Base.isapprox

miguelmaso:
Hi @Antoinemarteau , thanks for your review. I've applied your suggestions. However, you'll see that I've faced limitations when applying some of the suggestions. The PR is ready to review again.

JordiManyer:
I am just not sure why we are not just adding the kwargs to the pre-existing signature, instead of having two different ones? 

Like why not just change this: 

```julia
(≈)(a::MultiValue{S},b::MultiValue{S}) where {S} = isapprox(get_array(a), get_array(b))
(≈)(a::MultiValue{S,T1,N,0} where T1,b::MultiValue{S,T2,N,0} where T2) where {S,N} = true
```

to this: 

```julia
(≈)(a::MultiValue{S},b::MultiValue{S}; kwargs...) where {S} = isapprox(get_array(a), get_array(b); kwargs...)
(≈)(a::MultiValue{S,T1,N,0} where T1,b::MultiValue{S,T2,N,0} where T2; kwargs...) where {S,N} = true
```

or am I missing something? 

In my mind there is no need to have one signature with kwargs and one without.

Antoinemarteau:
I agree @JordiManyer, I suggested replacing the line cause I liked removing `{S}` from the signature because `isapprox([1 2], [1 2 3])` throws a `DimensionMismatch` error.
If we keep `{S}`, we get a method error for missmatching `::MultiValue` sizes, otherwise the next dispatch falls back to the mismatch error.

So I would keep
```julia
(≈)(a::MultiValue,b::MultiValue; kwargs...) = isapprox(get_array(a), get_array(b); kwargs...)
(≈)(a::MultiValue{S,T1,N,0} where T1,b::MultiValue{S,T2,N,0} where T2; kwargs...) where {S,N} = true
```

JordiManyer:
Sure, I think we can even remove the zero-dimensional dispatch and just keep the first signature, right? I don't think it gets used much and even if it does I presume the fallback is just as fast?

miguelmaso:
> I am just not sure why we are not just adding the kwargs to the pre-existing signature, instead of having two different ones?
> 
> Like why not just change this:
> 
> ```julia
> (≈)(a::MultiValue{S},b::MultiValue{S}) where {S} = isapprox(get_array(a), get_array(b))
> (≈)(a::MultiValue{S,T1,N,0} where T1,b::MultiValue{S,T2,N,0} where T2) where {S,N} = true
> ```
> 
> to this:
> 
> ```julia
> (≈)(a::MultiValue{S},b::MultiValue{S}; kwargs...) where {S} = isapprox(get_array(a), get_array(b); kwargs...)
> (≈)(a::MultiValue{S,T1,N,0} where T1,b::MultiValue{S,T2,N,0} where T2; kwargs...) where {S,N} = true
> ```
> 
> or am I missing something?
> 
> In my mind there is no need to have one signature with kwargs and one without.

I'm afraid the binary operator `≈` takes only two arguments. From the [documentation](https://docs.julialang.org/en/v1/base/math/#Base.isapprox):

> The binary operator `≈` is equivalent to `isapprox` with the default arguments

JordiManyer:
Then lets keep 

```julia
isapprox(a::MultiValue,b::MultiValue;kwargs...) = isapprox(get_array(a),get_array(b);kwargs...)
```

and remove the rest.

miguelmaso:
> Then lets keep
> 
> ```julia
> isapprox(a::MultiValue,b::MultiValue;kwargs...) = isapprox(get_array(a),get_array(b);kwargs...)
> ```
> 
> and remove the rest.

Got it. Done

Antoinemarteau:
> I'm afraid the binary operator `≈` takes only two arguments. From the [documentation](https://docs.julialang.org/en/v1/base/math/#Base.isapprox):

Its a minor detail, but this doesn't mean we need to overload `isapprox`. The "binary operator" refers to the infix notation `a ≈ b`. But `≈` is also an alias to `isapprox`. 
image
So it is not needed to import the `isapprox` symbol, importing and defining methods on `≈` is enough.
(The only reason why `a isapprox b` doesn't work is because of the parsing).

I also agree that we can remove the signature for empty tensors, the new default works the same.

miguelmaso:
> > I'm afraid the binary operator `≈` takes only two arguments. From the [documentation](https://docs.julialang.org/en/v1/base/math/#Base.isapprox):
> 
> Its a minor detail, but this doesn't mean we need to overload `isapprox`. The "binary operator" refers to the infix notation `a ≈ b`. But `≈` is also an alias to `isapprox`. image So it is not needed to import the `isapprox` symbol, importing and defining methods on `≈` is enough. (The only reason why `a isapprox b` doesn't work is because of the parsing).

Thanks for the explanation. Done

miguelmaso:
> I also agree that we can remove the signature for empty tensors, the new default works the same.

Removed line 8

miguelmaso:
Thank you very much for your attention!

Issue #1160, opened 2025-09-10
Accessing tensors by blocks or slices


miguelmaso:
**Introduction**
When computing stressess in a Gauss point for a visco-elastic model, I need to solve a system of equations. This system involves state variables and a lagrange multiplier. For instance, when computing the residual of this system, I build the full residual and its derivatives from the tensors corresponding to the different variables:
```julia
    res = [get_array(res1)[:]; res2]
    ∂res = MMatrix{10,10}(zeros(10, 10))
    ∂res[1:9, 1:9] = get_array(∂res1_∂Ce)
    ∂res[1:9, 10] = get_array(∂res1_∂λα)[:]
    ∂res[10, 1:9] = (get_array(∂res2_∂Ce)[:])'
```

There are lots of conversions between `StaticArray` and `TensorValue`. In other parts of the code I'm doing tones of conversions due to this limitation, e.g.:
```julia
Ce += TensorValue{3,3}(Tuple(Δu[1:end-1]))
```
or
```julia
VectorValue([get_array(Uv)[:]; λα])
```

**Proposal**
It'd be nice if it's possible to extend the `getindex` and `setindex!` functions.

I think botth additions need to be discussed since they involve design decisions which are out of my scope. Especially, the last function needs further considerations since TensorValue is inmutable.

**Discussion about `getindex`**
In PR #1159 there is a draft implementation for `getindex`. Some examples can be found in the tests.

My biggest question is about the return type for the possible combination of types and operators. However, it should be considered in combination with the potential assignment methods (e.g. setindex), which should be also discussed.

The first option is to always return the underlying `data`, so there is no need to care about the shape of the return type.

Option A | [ : ] | [ : , : ] | [ : , : , : ]
-|:-:|:-:|:-:
MultiValue | NTuple | ❌ | ❌
VectorValue | NTuple | ❌ | ❌
TensorValue | NTuple | NTuple | ❌
ThirdOrder | NTuple | ❌ | NTuple

The second option is to return a `MultiValue`, inspired in the behaviour of `StaticArrays`.

Option B | [ : ] | [ : , : ] | [ : , : , : ]
-|:-:|:-:|:-:
MultiValue | VectorValue | ❌ | ❌
VectorValue | VectorValue | ❌ | ❌
TensorValue | VectorValue | TensorValue/VectorValue | ❌
ThirdOrder | VectorValue | ❌ | ThirdOrder/TensorValue/VectorValue

[...]

PR #1161, opened 2025-09-10
Extended multi-indexing with slices - Return NTuple

miguelmaso:
Just an example for issue #1160


miguelmaso:
Closing as it is not needed based on #1160

PR #1173, opened 2025-10-13
[moment-based-reffes] Geometric decomposition on n-cubes, choice of poly basis in Lagrangian and Serendipity FES

Antoinemarteau:
Following my experiments on polynomial bases, I changed the default bases for moment elements on n-cubes and added support for  `poly_type` keywords for Lagrangian and Serendipity elements. Also implemented moment-based scalar reffes.

- generalized the `get_orders(::PolynomialBasis)` API to all polynomial bases, based on its existing behavior on `MonomialBasis` (for multi-valued bases, the order in each direction is the maximum of any component)
- Added `poly_type` keywords to `LagrangianRefFE` and `SerendipityRefFE`, but defaults is still `Monomial` because this only becomes useful at very high order (>5) and RefFE used for geometrical maps need to be monomials.
- Implemented geometric decomposition for Q⁻Λᵏ   `ModalC0` and `Bernstein` polynomial bases, and SΛ⁰ `ModalC0` bases
- Implemented moment-based scalar FE  for spaces {P⁻,P,Q⁻,S}Λ⁰ under the name `ModalScalar`
- Set `poly_type=ModalC0` as new default for moment-based RefFEs on n-cubes (instead of Legendre), along with `sh_is_pb=true` by default
- Added `mom_poly_type` keyword to moment-based RefFEs, giving choice for the basis of the moment test spaces. Defaults to `poly_type` (best according to my experiments)
- Added `nodal=false` keyword argument to the `ReferenceFE` constructor using FEEC notation, for 0-forms (scalar functions), this enables opting-in for nodal DOF bases (translating to `lagrangian` and `serendipity` reffes) instead of the default moment-based DOF bases (`ModalScalar` reffes)
- added support to Cartesian product space to all scalar reffes constructors (this was lacking in the FEEC notation based constructors)

### DIFF
--- a/docs/src/modules/ReferenceFEs.md
+++ b/docs/src/modules/ReferenceFEs.md
@@ -46,7 +46,7 @@ The following table summarizes the elements implemented in Gridap (legend below)
 |                                                                                         |                                              | 𝓟ᵣΛᴰ      | △           | ``{r=o≥0, o}``  | `:L2`     |
 |                                                                                         |                                              | 𝓠ᵣ⁻Λᴰ     | ``\square`` | ``{r-1=o≥0, o}``| `:L2`     |
 |                                                                                         | kwarg `space=:P`                             | 𝓢ᵣΛᴰ      | ``\square`` | ``{r=o≥0, o}``  | `:L2`     |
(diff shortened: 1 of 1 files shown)


Antoinemarteau:
Edit: I will have to remove `sh_is_pb` keyword from `lagrangian` reffes because it doesn't make sense.

PR #1184, opened 2025-11-19
[Moment based reffes] small TensorValues and Polynomials improvements, MomentDofBasis generalization to differential operators

Antoinemarteau:
This mostly prepares the ground for C1 refference FES, but is ready to merge now.

Additions:
- Added `representatives_of_basis_dual`, same as `representatives_of_componentbasis_dual` but takes a basis instead of a tensor type (to compute dual to subpaces of span{component_basis(V)})
- implemented `vcat(::AbstractVector{<:Dof}...)` via `ConcatenatedDofVector`, we can now concatenate arbitrary dof bases. The `face_own_dof` concatenation has to be done by hand independantly becase `LagrangianDofBasis` does not store it's  face ownership nor underlying polytope.
- Generalized `MomentBasedDofBasis` to enable precomposition of the given field with a function, e.g. a differential operator. The selection of e.g. gradient component or application of directional derivative is delegated to the given moment functional and test basis.

Modified:
- refactor and generalized derivative evaluation of `CartProdPolyBasis` (including `MonomialBasis`) to tensor valued polynomial with any dependant components pattern
- replaced the `face_dofs` field of `GenericReferenceFE` with `face_own_dofs`, it is more consistent with what we use and enables a generic default implementation for `get_face_dofs`.
- deprecated `return_type(::PolynomialBasis)` in favor of `value_type(::PolynomialBasis)`
- minor docs improvements

### DIFF
--- a/src/Polynomials/PolynomialInterfaces.jl
+++ b/src/Polynomials/PolynomialInterfaces.jl
@@ -87,6 +87,7 @@ For tensor-valued bases, it is the maximum for each component.
 """
 @inline get_order(::PolynomialBasis) = @abstractmethod
 get_order(f::LinearCombinationFieldVector) = get_order(f.fields)
+get_order(f::FieldGradientArray) = get_order(f.fa)
 get_order(f::AbstractVector{<:ConstantField}) = 0
 
 """
--- a/src/ReferenceFEs/MomentBasedReferenceFEs.jl
+++ b/src/ReferenceFEs/MomentBasedReferenceFEs.jl
@@ -4,21 +4,27 @@
 struct Moment <: Dof end
 
 """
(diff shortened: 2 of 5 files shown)


Antoinemarteau:
Ok, I investigated these broken Adaptivity tests for 2h.

I believe they are not breaking due to the changes on these PR, I can reproduce them on Gridap.jl/moment-based-reffes (before this PR) by updating the package with latest registeries, which the CI apparently does (it at least re-resolve) ?

So I assume we need to restrict the responsible dependency in Project.toml, for it not to update too much? 
I don't know which one is responsible though... updating breaks current `moment-based-reffes` but not `master`.

Possibly suspicious updates I see are  `ForwardDiff v0.10.38 ⇒ v1.3.0`, `BlockArrays v1.2.0 ⇒ v1.9.2`

PR #1212, opened 2026-01-28
Added support for `sqrt`

miguelmaso:
Hi *Gridap* maintainers, I'm developing some constitutive models in another [package](https://github.com/MultiSimOLab/HyperFEM) (not public yet). For convenience, I've defined the `sqrt` of a `TensorValue`, but I'd like to have its definition inside *Gridap* to avoid type piracy.

There are some considerations below, I'm totally open to modify this implementation according to your code guidelines.

## Some considerations:

**Efficiency vs maintainability**: the computation of the square root is delegated to `StaticArrays`. There could be more efficient implementations (especially for 3x3 matrices and assuming symmetric positive definite case), but I find this implementation easier to maintain.

**Docstrings**: I havn't added docstrings because I find them redundant with those that come from `Base`.

## See also: efficiency

This is the code I've used to test efficiency:
```julia
using LinearAlgebra, StaticArrays, Gridap, BenchmarkTools, Plots
import Base:sqrt

sqrt(a::TensorValue{D}) where D = TensorValue{D}(sqrt(get_array(a)))

function msqrt(A::TensorValue{D}) where D
  λ, v = eigen(A)
  λ = sqrt.(λ)
  TensorValue{D}(v*diagm(λ)*v')
end

n = 10
time_ssqrt = zeros(n)
time_msqrt = zeros(n)
sizes = [1:n...]

for i ∈ sizes
  a = SMatrix{i,i}(rand(i,i))
  a = a*a'
  b = TensorValue(a)
  time_ssqrt[i] = @belapsed sqrt($a)
  time_msqrt[i] = @belapsed msqrt($b)
end

plot(sizes, [time_ssqrt, time_msqrt], labels=["Static" "eigen"])
```

image

### DIFF
--- a/src/TensorValues/Operations.jl
+++ b/src/TensorValues/Operations.jl
@@ -761,7 +761,7 @@ Normalization of a tensor value.
 """
 normalize(a::MultiValue) = a / norm(a)
 
-sqrt(a::TensorValue{D}) where D = TensorValue{D}(sqrt(get_array(a)))
+sqrt(t::MultiValue{Tuple{D,D}}) where D = TensorValue{D,D}(sqrt(SArray(t)))
 
 ###############################################################
 # Measure
--- a/test/TensorValuesTests/OperationsTests.jl
+++ b/test/TensorValuesTests/OperationsTests.jl
@@ -664,6 +664,16 @@ expected3 = eigen(get_array(t3))
 t_nonsquare = TensorValue{2,3}(1:6...)
 @test_throws ErrorException eigen(t_nonsquare)
 
+# Test sqrt function
+t = TensorValue(4.0,0.0,0.0,4.0)
+@test sqrt(t) == TensorValue(2.0,0.0,0.0,2.0)
+t = SymTensorValue{2}(1,0,0)
+@test sqrt(t) == TensorValue(1.0,0.0,0.0,0.0)
+t = SymTracelessTensorValue{2,ComplexF64}(1,0)
+@test sqrt(t) == TensorValue(1.0, 0.0, 0.0, -im)
+t = SymTensorValue{2}(-1,0,0)
+@test_throws DomainError sqrt(t)
+
 # Measure
 
 a = VectorValue(1,2,3)


JordiManyer:
Hi @miguelmaso thank you for your contribution! I agree that delegating to static arrays is more maintainable. If it really makes a difference computationally, we can maybe add hard-coded signatures for 2x2 and 3x3 tensors, but its up to you. 

Add a couple tests so that codecov is happy, and I'll merge when you tell me.

Antoinemarteau:
It would be nice to implement it for the other 2x2 tensor types:
```julia
sqrt(t::SymTensorValue{D}) where D = SymTensorValue{D}(sqrt(SArray(t)))
sqrt(t::SkewSymTensorValue{D}) where D = TensorValue{D,D}(sqrt(SArray(t)))
sqrt(t::SymTracelessTensorValue{D}) where D = TensorValue{D,D}(sqrt(SArray(t)))
```
( I'm not 100% that the sqrt of a symetric tensor is always symetric, is it? )

Matrices for tests:
```julia
t = SymTensorValue{2}(0,1,0)
t = SkewSymTensorValue{2}(1)
t = SymTracelessTensorValue{2,ComplexF64}(1,2)
```

you can also factorize the three signature for `TensorValue`, `SkewSymTensorValue`, and `SymTracelessTensorValue` by:
```julia
sqrt(t::MultiValue{Tuple{D,D}}) where D = TensorValue{D,D}(sqrt(SArray(t)))
```
(but keep that for `SymTensorValue` that has different return type)

miguelmaso:
Hello @JordiManyer and @Antoinemarteau , thanks for your comments.


> ( I'm not 100% that the sqrt of a symetric tensor is always symetric, is it? )

The sqrt of a symmetric tensor is symmetric if all the eigenvalues are ≥ 0 (positive semidefinite - PSD). Otherwise, it's non-symmetric or complex (there is no unique sqrt). Hence, the implementation
```jl
sqrt(t::SymTensorValue{D}) where D = SymTensorValue{D}(sqrt(SArray(t)))
```
is valid as long as `t` is PSD. In my case, it holds, but there could be someone using non-PSD matrices.

As @Antoinemarteau proposes, I'd go only for
```jl
sqrt(t::MultiValue{Tuple{D,D}}) where D = TensorValue{D,D}(sqrt(SArray(t)))
```
because it is the most generic case and specializations are delegated to *StaticArrays*. Of course, a dispatch for `SymTensorValue` can be added, but the decision is up to you: generic vs specialized.

> Matrices for tests:
> 
> ```julia
> t = SymTensorValue{2}(0,1,0)
> t = SkewSymTensorValue{2}(1)
> t = SymTracelessTensorValue{2,ComplexF64}(1,2)
> ```

I'll take care about it.

Antoinemarteau:
Ok if it's not always symmetric don't keep a specific method for `SymTensorValue`

miguelmaso:
Hello, sorry for the delay. It's ready to be merged:

- [x] Added signature for `TensorValue`, `SymTensorValue`, and `SymTracelessTensorValue`
- [x] Added entry in `News.md`
- [x] Added tests

Let's wait for the CI

miguelmaso:
Thanks!

PR #1213, opened 2026-02-02
fix and refactor Arrays.getindex!

Antoinemarteau:
`getindex!` would ignore the cache if the given indices did not match the `IndexStyle` of the array. This could lead to re-alocating the cache of `LazyArray`s at each access in hot loops.

Now, the indices are converted using LinearIndices or CartesianIndices, so the cache should never be ignored if the indices index a single element in the array.

This also cleans the implementation by using @propagate_inbounds, limiting number of method and using Base to transform the indices.


Antoinemarteau:
I also took the opportunity to add `@Base.propagate_inbounds` to loads of `getindex!` / `setindex!` methods in `Arrays`, and add `@inbounds` where I felt it was safe in the module.

PR #1216, opened 2026-02-07
Add nedelec test facet dofs normals

amartinhuertas:


Antoinemarteau:
Rq, the added test passes on the `moment-based-reffes` branch, if using the Berstein basis (which is the default there). 
The test is still broken if forcing using monomial basis in the reffe. 
I assume it's because the Bernstein bases are "symmetric" with respect to barycentric coordinates and local indices, but I don't know if that fix would remain for higher orders.

Issue #1221, opened 2026-02-19
Incorrect Jacobian indexing in Nedelec DOF basis transformation


Ady0333:
### **Describtion** 
In `TransformNedelecDofBasis.evaluate!` (`src/FESpaces/CurlConformingFESpaces.jl`), the Jacobian array `Jtx` is indexed using the local loop index p instead of the global node index from `dofs.face_nodes[face]`.

Current code:
```julia
F = transpose(Jtx[p])
```
But p is only a local index on the face, while Jtx is indexed globally over all DOF quadrature points in the cell.

For faces beyond the first, this ends up using the Jacobian from a wrong quadrature point during the Piola transform of the DOF moments.

---

### **Expected behavior**
The Jacobian should be accessed using the corresponding global node index:
```julia
F = transpose(Jtx[face_point_ids[p]])
```
---

### **Impact**
This affects H(curl) (Nedelec) elements on non-affine meshes (e.g. quads/hexes or curved geometries), where the Jacobian varies within the cell.

It can lead to incorrect DOF basis transformation and wrong assembled system matrices without any runtime error.

---
### **Affected File**
`src/FESpaces/CurlConformingFESpaces.jl`

Antoinemarteau:
Hey @Ady0333, could you add a test relevant to the issue ? 
We are actually refactoring the Piola maps machineries so that would help validating the new implementation!

Ady0333:
> Hey [@Ady0333](https://github.com/Ady0333), could you add a test relevant to the issue ? We are actually refactoring the Piola maps machineries so that would help validating the new implementation!

Yes sir sure!! I'll add it as soon as possible.

PR #1222, opened 2026-02-19
Fix Nedelec DOF basis Jacobian indexing on multi-face elements

Ady0333:
This PR fixes an indexing issue in the DOF basis transformation for Nedelec (H(curl)) finite elements reported in #1221.

In `TransformNedelecDofBasis.evaluate!` (`CurlConformingFESpaces.jl`), the Jacobian transpose array `Jtx` was being accessed using the local loop index `p`:

```julia
F = transpose(Jtx[p])
```

However, the corresponding quadrature points for each face are indexed globally through:

```julia
face_point_ids = dofs.face_nodes[face]
```

Since `face_point_ids` stores global node indices, using the local index `p` may result in accessing the Jacobian evaluated at an incorrect quadrature point.

For the first face this goes unnoticed because the range starts at 1 (i.e. `p == face_point_ids[p]`). For all subsequent faces, the range starts at an offset, causing Jacobian values to be read from incorrect locations during the Piola transformation of the DOF moments.

---

### Impact

This affects H(curl) problems using Nedelec elements on meshes with non-affine geometric mappings (e.g. quadrilateral/hexahedral meshes or curved elements), where the Jacobian varies within the cell.

Incorrect indexing during DOF basis transformation may lead to wrongly assembled system matrices without producing any runtime error.

---

### Fix

Use the corresponding global node index when accessing the Jacobian:

```julia
F = transpose(Jtx[face_point_ids[p]])
```

This matches the same indexing pattern already used in the RT (H(div)) implementation in `DivConformingFESpaces.jl`.

### DIFF
--- a/test/FESpacesTests/CurlConformingFESpacesTests.jl
+++ b/test/FESpacesTests/CurlConformingFESpacesTests.jl
@@ -209,23 +209,28 @@ eh = fh - f
 #using Gridap.Visualization
 #writevtk(trian,"trian",nsubcells=10,cellfields=["uh"=>uh])
 
-# Regression test for PR #1222: TransformNedelecDofBasis.evaluate! used
-# Jtx[p] instead of Jtx[face_point_ids[p]], causing wrong Jacobian indexing.
-@testset "Regression #1222: TransformNedelecDofBasis Jacobian indexing" begin
+# Regression test for PR #1222: TransformNedelecDofBasis.evaluate! must use
+# Jtx[face_point_ids[p]], not Jtx[p].  Affine cells (rectangular quads or
+# simplices) have a constant Jacobian, so both indices coincide and the bug is
+# latent.  A bilinear (non-affine) quad mesh makes the Jacobian vary across
+# DOF nodes of different edges, exposing the wrong-index path.  The constant
+# field (2,3) is exactly representable in the physical Nedelec space for any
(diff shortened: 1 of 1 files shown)


JordiManyer:
@tamaratambyah @amartinhuertas I think this might be somehow related to your issues? I leave the review to you!

JordiManyer:
@Antoinemarteau

Antoinemarteau:
We need a test for this @Ady0333 in order to merge.

Ady0333:
> We need a test for this @Ady0333 in order to merge.

Okay @Antoinemarteau i will add a test shortly.

Apologies for late communication there was some technical problems due to which i was not able to communicate.

JordiManyer:
@Ady0333 the test does not fail for Gridap#master. You are not targeting the actual bug, which only shows for non-affine geometric maps. This is the problem when you have no idea of the maths and just rely on llms.

Ady0333:
@JordiManyer 
Thanks for pointing that out, and sorry about the confusion.

You're right , my test does not actually exercise the non-affine geometric mapping case where the bug manifests. I overlooked that the issue only appears when the Jacobian varies within the cell. I’ll revisit the test and construct one that explicitly uses a non-affine map so it targets the correct failure mode.

Thanks for the guidance.

Issue #1223, opened 2026-02-20
Incorrect inverse cell map indexing in FineToCoarseField fast-path


Ady0333:
## Describe the bug

In the fast-path implementation of `FineToCoarseField.evaluate!` (`src/Adaptivity/FineToCoarseFields.jl`), the inverse cell map array `cmaps` is indexed using `field_id` instead of `child_id`.

Current code:

```julia
mi = getindex!(mi_cache, cmaps, field_id)
```

However, `cmaps` is indexed by subcell ID, and `field_id` may not match the corresponding `child_id` when only a subset of children is locally present (e.g. in distributed adaptive meshes). This can result in using the inverse map of a different child cell during field evaluation and may lead to incorrect DOF evaluation on adapted meshes (for example in distributed adaptive simulations using GridapP4est) without producing any runtime error.

---

### Impact

This may lead to incorrect DOF evaluation on adapted meshes (for example in distributed adaptive simulations using GridapP4est) without producing any runtime error.

---

## Affected file

```
src/Adaptivity/FineToCoarseFields.jl
```

PR #1224, opened 2026-02-20
Fix/ftc field cmaps index

Ady0333:
### Description

This PR fixes incorrect inverse cell map indexing in the fast-path implementation of `FineToCoarseField.evaluate!` reported in #1223.

In `src/Adaptivity/FineToCoarseFields.jl` (line 150), the inverse cell map array `cmaps` was being indexed using `field_id`:

```julia
mi = getindex!(mi_cache, cmaps, field_id)
```

However, `cmaps = get_inverse_cell_map(rr)` contains one map per subcell of the refinement rule and must be indexed using `child_id`. `field_id` is the index into `fine_fields`, which may differ from the subcell index when not all children are locally present.

This PR updates the fast-path evaluation to use the correct index:

```julia
mi = getindex!(mi_cache, cmaps, child_id)
```

---

### Impact

This affects DOF evaluation on adapted meshes when only a subset of child cells is available (e.g. in distributed adaptive simulations using GridapP4est). Using the wrong inverse map during field evaluation can lead to incorrect DOF values without producing runtime errors.

---

### Fix

Use `child_id` instead of `field_id` when indexing `cmaps` in the fast-path evaluation. This matches the same fix already applied to the slow path in commit 76901ef1b.

### DIFF
--- a/src/Adaptivity/FineToCoarseFields.jl
+++ b/src/Adaptivity/FineToCoarseFields.jl
@@ -147,7 +147,7 @@ function Arrays.evaluate!(cache,a::FineToCoarseField,x::AbstractArray{<:Point},c
     field_id = id_map[child_id]
     if field_id != 0
       fi = getindex!(fi_cache,fields,field_id)
-      mi = getindex!(mi_cache,cmaps,field_id)
+      mi = getindex!(mi_cache,cmaps,child_id)
       zi = Fields.evaluate!(zi_cache,mi,xi)
       y[i] = Fields.evaluate!(yi_cache,fi,zi)
     end


JordiManyer:
@Antoinemarteau

Antoinemarteau:
By comparison with #1074, this seems correct. I struggle to see how to adapt the test in #1074 to generalize it to the method using `child_ids` in argument.
If a test isn't needed this can be merged.

Ady0333:
@Antoinemarteau 
Thanks for checking!

I also looked at the test added in #1074, but I couldn’t find a straightforward way to adapt it to cover the `child_ids` evaluation path without introducing additional setup.

If you think a dedicated test is not necessary here, I'm happy to leave it as is. Otherwise I can try to explore a more specific test for the fast-path evaluation.

Issue #1225, opened 2026-02-22
MappedGrid does not override get_cell_map


Ady0333:
## Describe the bug

`MappedGrid` computes and stores a composed geometric map

```
geo_map = phys_map ∘ model_map
```

in its constructor, but it does not override `get_cell_map`.

Because of this, calling `get_cell_map` on a `MappedGrid` falls back to the default implementation for `Grid`, which reconstructs the map using linear interpolation of the deformed node coordinates.

```julia
function get_cell_map(trian::Grid)
  cell_to_coords = get_cell_coordinates(trian)
  cell_to_shapefuns = get_cell_shapefuns(trian)
  lazy_map(linear_combination, cell_to_coords, cell_to_shapefuns)
end
```

This reconstruction ignores the stored `geo_map`.

---

## Impact

For nonlinear or curved `phys_map`, the reconstructed map becomes a piecewise linear approximation of the geometry instead of the intended composed map. This can lead to incorrect Jacobians and integration results during FE assembly.

---

## Affected file

```
src/Geometry/MappedDiscreteModels.jl
```

---

## Possible fix

Return the stored map from `MappedGrid`:

```julia
get_cell_map(grid::MappedGrid) = grid.geo_map
```

Ady0333:
Please assign me this issue.

PR #1226, opened 2026-02-22
Fix/issue 1224 mapped grid cell map

Ady0333:
This PR fixes the issue reported in #1225 where `MappedGrid` does not override `get_cell_map`.

`MappedGrid` computes and stores the composed geometric map

```
geo_map = phys_map ∘ model_map
```

during construction. However, there was no `get_cell_map(grid::MappedGrid)` method returning this map. As a result, calls to `get_cell_map` fell back to the default implementation for `Grid`, which reconstructs the mapping using linear interpolation of the deformed node coordinates.

```julia
function get_cell_map(trian::Grid)
  cell_to_coords = get_cell_coordinates(trian)
  cell_to_shapefuns = get_cell_shapefuns(trian)
  lazy_map(linear_combination, cell_to_coords, cell_to_shapefuns)
end
```

This reconstruction is correct only when the physical mapping is linear. For nonlinear or curved mappings it produces a piecewise linear approximation instead of the composed map stored in `MappedGrid`.

---

### Impact

For grids created with a nonlinear `phys_map`, the fallback behavior may lead to incorrect geometric mappings during FE assembly. This affects quantities derived from the mapping such as Jacobians, integration weights, and gradient transformations.

As a result, FE operators assembled on a `MappedGrid` may use an approximate geometric map instead of the intended composed map.

---

### Fix

Return the stored composed map when `get_cell_map` is called on a `MappedGrid`:

```julia
get_cell_map(grid::MappedGrid) = grid.geo_map
```

---

### Summary of changes

- Added `get_cell_map(::MappedGrid)` in  
  `src/Geometry/MappedDiscreteModels.jl`
- The method returns the stored composed map (`geo_map`)
- Ensures `MappedGrid` provides the correct geometric mapping during FE computations

### DIFF
--- a/src/Geometry/MappedDiscreteModels.jl
+++ b/src/Geometry/MappedDiscreteModels.jl
@@ -55,6 +55,7 @@ get_node_coordinates(grid::MappedGrid) = grid.node_coords
 get_cell_node_ids(grid::MappedGrid) = get_cell_node_ids(grid.grid)
 get_reffes(grid::MappedGrid) = get_reffes(grid.grid)
 get_cell_type(grid::MappedGrid) = get_cell_type(grid.grid)
+get_cell_map(grid::MappedGrid) = grid.geo_map
 
 """
 MappedDiscreteModel


Ady0333:
@JordiManyer Please review this pr.

JordiManyer:
@Antoinemarteau

Antoinemarteau:
Are you actually using `MappedGrid` @Ady0333 ? 

I think this code is an unfinished experiment that is actually unused, it would probably require more work than this PR to be useful.

Ady0333:
@Antoinemarteau Thanks for the review!

I’m not currently using `MappedGrid` directly. I noticed the issue while exploring the geometry code and saw that `geo_map` was computed but never returned by `get_cell_map`.

My intention with this PR was just to make the stored map accessible if `MappedGrid` is used, but I understand your point that the feature might be incomplete or experimental.

If this part of the code is currently unused or planned to be reworked, I'm happy to close the PR or adjust it depending on what you think would be most useful.

JordiManyer:
I think the correctness of this PR is debatable. Right now, the cell map that is returned is the polynomial approximation of the internal map `geo_map`, where the polynomial order depends on the grid. One could argue this is not the intended behaviour, and that we should indeed be returning the internal variable (which could contain a nonlinear map, etc...). 
As per the description, "it can be used to include a high order map implemented by any map that is
 a `CellField`", which I would say points to this fix being correct. 

In any case, I would agree with @Antoinemarteau  that the `Mapped...` structs need to be reworked to be actually useful. But I think this is a step in the right direction, probably. What do you think @Antoinemarteau ?

Antoinemarteau:
I think that these `Mapped...` struct will require a complete review and concrete use case tests to come alive, so this PR has not much effect. I think its both fine to merge or not, and we should just close this ASAP..

PR #1232, opened 2026-02-25
Fix/block multifield permutation offsets

anusha19murthy:
This PR fixes the issue where `compute_field_offsets` in `MultiFieldFESpaces.jl` applies the forward permutation instead of the inverse permutation when recovering per-field offsets for `BlockMultiFieldStyle`.

`compute_field_offsets` constructs `block_offsets` following the permuted block layout order. However, the returned offsets must correspond to the original field ordering. The current implementation incorrectly applies the forward permutation:

```julia
offsets = map(p -> block_offsets[p], P)
```

Since `block_offsets` is already ordered according to the permuted block structure, recovering offsets per original field requires applying the inverse permutation.

Using the forward permutation silently assigns offsets to the wrong fields for non-involutive permutations (e.g. P = (3,1,2)), resulting in swapped or overlapping field DOF slices.

**Impact**

For `MultiFieldFESpace` instances using `BlockMultiFieldStyle` with explicit permutations:

- `_restrict_to_field` extracts incorrect DOF slices

- field DOF ranges may be swapped between fields

- assembly and solves complete without errors

- resulting solutions silently assign DOFs to the wrong fields

This produces incorrect numerical results in multifield simulations while solver convergence appears normal.

**Fix**

Apply the inverse permutation when recovering offsets.

Replace:

`offsets = map(p -> block_offsets[p], P)`

with:

```
Pinv = invperm(P)
offsets = map(p -> block_offsets[p], Pinv)
```

This ensures offsets correspond to the correct original field ordering.

**Summary of changes**

Fixed permutation handling in `compute_field_offsets`

Applied inverse permutation instead of forward permutation

Prevented silent swapping of field DOF slices in permuted `BlockMultiFieldStyle`

Restored correct DOF extraction for multifield problems with non-identity permutations

### DIFF
--- a/test/MultiFieldTests/MultiFieldFESpacesTests.jl
+++ b/test/MultiFieldTests/MultiFieldFESpacesTests.jl
@@ -142,6 +142,30 @@ interpolate_everywhere!([f,f],x,dv,X)
   uh_perm, ph_perm = xh_perm
   @test sum(∫((uh_ref - uh_perm)*(uh_ref - uh_perm))dΩ2) < 1e-24
   @test sum(∫((ph_ref - ph_perm)*(ph_ref - ph_perm))dΩ2) < 1e-24
+  V1 = TestFESpace(model2,ReferenceFE(lagrangian,Float64,2);conformity=:H1)
+  V2 = TestFESpace(model2,ReferenceFE(lagrangian,Float64,1);conformity=:L2)
+  V3 = TestFESpace(model2,ReferenceFE(lagrangian,Float64,1);conformity=:L2)
+
+  # Self-inverse permutation P=(2,1): both fields alone in their blocks → offsets all zero.
+  # This case was unaffected by either bug.
+  mfs1 = BlockMultiFieldStyle(2,(1,1),(2,1))
+  Y1 = MultiFieldFESpace([V1,V2]; style=mfs1)
+  off1 = MultiField.compute_field_offsets(Y1)
+  @test off1[1] == 0
+  @test off1[2] == 0
+
+  # Non-self-inverse permutation P=(3,1,2), SB=(2,1):
+  #   block 1 = [field3, field1],  block 2 = [field2]
+  # Correct offsets (after fix):
+  #   field3 leads block1  → offset 0
+  #   field1 follows field3 in block1 → offset = num_free_dofs(V3)
+  #   field2 alone in block2 → offset 0
+  mfs2 = BlockMultiFieldStyle(2,(2,1),(3,1,2))
+  Y2 = MultiFieldFESpace([V1,V2,V3]; style=mfs2)
+  off2 = MultiField.compute_field_offsets(Y2)
+  @test off2[3] == 0
+  @test off2[1] == num_free_dofs(V3)
+  @test off2[2] == 0
 end
 
 end # module


anusha19murthy:
@Antoinemarteau Please do review this PR. Happy to make any changes if needed!

Antoinemarteau:
Hi @anusha19murthy, when you open a PR, please add a test, possibly a MWE (minimum working example), that passes after the change (but would fails before it).
This is necessary to understand the necessity of the change, and be sure to keep the fix if the code is modified again later.

anusha19murthy:
Hi @Antoinemarteau, thanks for the clarification.
Yes, that makes sense. I’ll add a minimal test case exercising BlockMultiFieldStyle with a non-identity permutation, showing the incorrect behavior before the fix and the corrected behavior after it.
I’ll update the PR shortly.

anusha19murthy:
Hello @Antoinemarteau I've added the test. Please go through it once and let me know if any changes are needed. Thank you.

JordiManyer:
Hi @anusha19murthy the fix is correct, thanks for spotting the bug. I prefer the following battery of tests: 

```julia
model = CartesianDiscreteModel((0,1,0,1),(2,2))
V1 = TestFESpace(model,ReferenceFE(lagrangian,Float64,1))
V2 = TestFESpace(model,ReferenceFE(lagrangian,Float64,2))
V3 = TestFESpace(model,ReferenceFE(lagrangian,Float64,3))

dΩ = Measure(Triangulation(model),2)

a((u1,u2,u3),(v1,v2,v3)) = ∫((u1+u2+u3)*(v1-v2+2*v3))dΩ

X1 = MultiFieldFESpace([V1,V2,V3];style=BlockMultiFieldStyle(2,(2,1),(1,2,3)))
X2 = MultiFieldFESpace([V1,V2,V3];style=BlockMultiFieldStyle(2,(2,1),(3,1,2)))
X3 = MultiFieldFESpace([V1,V2,V3];style=BlockMultiFieldStyle(2,(1,2),(1,3,2)))

ndofs = map(num_free_dofs,(V1,V2,V3))
@test MultiField.compute_field_offsets(X1) == [0, ndofs[1], 0]
@test MultiField.compute_field_offsets(X2) == [ndofs[3], 0, 0]
@test MultiField.compute_field_offsets(X3) == [0, ndofs[3], 0]

A1 = assemble_matrix(a,X1,X1)
A2 = assemble_matrix(a,X2,X2)
A3 = assemble_matrix(a,X3,X3)
@test blocksizes(A1,1) == blocksizes(A1,2) == [ndofs[1]+ndofs[2], ndofs[3]]
@test blocksizes(A2,1) == blocksizes(A2,2) == [ndofs[3]+ndofs[1], ndofs[2]]
@test blocksizes(A3,1) == blocksizes(A3,2) == [ndofs[1], ndofs[3]+ndofs[2]]
```

Also do not forget to update NEWS.md

JordiManyer:
@Ady0333 can you resubmit this PR with the changes that we have requested? I remember you posting the same issue and I dont think the original author is active anymore. Cheers.

Ady0333:
@JordiManyer 
Thankyou for this opportunity. I will resubmit this pr with the changes that have requested. 
Please allow me some time.

Issue #1233, opened 2026-02-26
get_cell_dof_ids skips non-zero offset for field 1 in BlockMultiFieldStyle with permuted blocks, causing silent DOF corruption


anusha19murthy:
**Describe the bug**

`get_cell_dof_ids` in `MultiFieldFESpaces.jl` incorrectly skips applying the DOF offset for field `1` when using `BlockMultiFieldStyle` with a non-identity permutation `P`.

The implementation assumes that field 1 always has zero offset:

if i == 1
  push!(active_block_data,cell_dofs_i)

This assumption is valid for:

`ConsecutiveMultiFieldStyle`

`BlockMultiFieldStyle with identity permutation`

However, for `BlockMultiFieldStyle` with a non-identity permutation, field `1` may not be first within its block, and therefore `offsets[1]` may be non-zero.

Since `compute_field_offsets` correctly computes within-block offsets accounting for permutation, skipping the offset application for `i == 1` causes incorrect DOF numbering.

As a consequence, DOF slices belonging to different fields may overlap or be swapped.

**Impact**

Field offsets are later used to extract each field’s DOF slice from the global block vector.

When offsets are incorrect:

- fields may read DOF ranges belonging to other fields
- DOF slices can overlap or become swapped 
- matrix and vector assembly complete without errors
- linear and nonlinear solvers converge normally
- resulting solutions silently contain corrupted or swapped field values

This produces incorrect numerical results in multiphysics simulations using `BlockMultiFieldStyle` with explicit permutations.

The failure is silent and difficult to detect because:

- no bounds errors occur
- matrix dimensions are correct
- solver convergence is unaffected

This can silently corrupt multiphysics block-preconditioned simulations (e.g., Stokes, Darcy, coupled problems).

**Affected file**
`src/MultiField/MultiFieldFESpaces.jl`

**Function:**

```
FESpaces.get_cell_dof_ids(
  f::MultiFieldFESpace,
  trian::Triangulation,
  ::Union{<:ConsecutiveMultiFieldStyle,<:BlockMultiFieldStyle}
)
```

**Related:**

```
compute_field_offsets(
  f::MultiFieldFESpace,
  ::BlockMultiFieldStyle{NB,SB,P}
)
```
**Exact problematic code**
```
for i in active_block_ids
  cell_dofs_i = get_cell_dof_ids(f.spaces[i],trian)
  if i == 1
    push!(active_block_data,cell_dofs_i)
  else
[...]

PR #1236, opened 2026-02-27
Fix DOF offset computation in BlockMultiFieldStyle with non-identity …

anusha19murthy:
This PR fixes issue #1233 where `BlockMultiFieldStyle` with non-identity permutation `P` produces incorrect DOF mappings due to two independent bugs in `MultiFieldFESpaces.jl`.

**Describe the bug**
**Bug 1** — `compute_field_offsets` uses forward permutation
`offsets = map(p -> block_offsets[p], P)`

`block_offsets` is ordered by permuted block position.
To recover offsets per original field, the inverse permutation must be used.
Using `P` incorrectly reassigns offsets between fields. This was only correct for self-inverse permutations (identity, simple swaps), which is why existing tests passed.

**Bug 2** — `get_cell_dof_ids` assumes field 1 has zero offset
`if i == 1`

This assumes `offsets[1] == 0`, which is not true for non-identity permutations where field 1 is not first in its block. The offset is skipped, causing DOF overlap.

---

**Impact**

For non-identity permutations:

- DOF slices can overlap or be swapped
- Assembly completes without error
- Solutions are silently corrupted

This affects multiphysics simulations using explicit block permutations.

---

**Fix**

1. Use inverse permutation:
`offsets = map(p -> block_offsets[p], invperm([P...]))`

2. Apply offset based on value, not index:
`if offsets[i] == 0`

---

**Summary of changes**

- Corrected offset recovery in `compute_field_offsets`
- Fixed DOF offset application in `get_cell_dof_ids`
- Restored correct DOF numbering for all `BlockMultiFieldStyle` permutations
- Eliminated silent assembly corruption for non-identity `P`

### DIFF
--- a/test/MultiFieldTests/MultiFieldFESpacesTests.jl
+++ b/test/MultiFieldTests/MultiFieldFESpacesTests.jl
@@ -116,4 +116,32 @@ dv = zero_dirichlet_values(X)
 interpolate_everywhere!([f,f],x,dv,X)
 @test x == get_free_dof_values(fh)
 
+@testset "BlockMultiFieldStyle with non-identity permutation" begin
+  model2 = CartesianDiscreteModel((0,1,0,1),(2,2))
+  V1 = TestFESpace(model2,ReferenceFE(lagrangian,Float64,2);conformity=:H1)
+  V2 = TestFESpace(model2,ReferenceFE(lagrangian,Float64,1);conformity=:L2)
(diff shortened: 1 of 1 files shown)


JordiManyer:
Add another duplicated AI PR again, and I will automatically remote anything you do in this repository. You have been warned.

PR #1240, opened 2026-03-05
Fix outer(f, s::ShiftedNabla) ignoring shift term

Ady0333:
This PR fixes the issue #1241 where `outer(f, s::ShiftedNabla)` ignored the shift term and returned `transpose(∇f)` instead of the shifted operator.

`ShiftedNabla` represents the differential operator `∇ + v⊗`. Applying it to a field `f` yields:

```
s(f) = ∇f + v ⊗ f
```

However, the implementation of

```
outer(f, s::ShiftedNabla)
```

incorrectly used:

```
transpose(gradient(f))
```

which corresponds to the plain gradient operator and silently discards the `v ⊗ f` contribution.

This line was copied from the implementation for the standard gradient case:

```
outer(f, ::typeof(∇)) = transpose(gradient(f))
```

but was never updated to apply the shift term.

As a result, expressions involving `f ⊗ (∇ ± v)` evaluated as `transpose(∇f)` instead of the expected shifted operator.

---

## Impact

When `ShiftedNabla` appears on the right-hand side of `outer`, the shift term `v ⊗ f` is silently ignored.

This produces mathematically incorrect expressions in weak forms involving shifted gradient operators, since

```
outer(f, s) ≠ transpose(outer(s, f))
```

even though the operator algebra should satisfy this identity.

As a consequence, formulations that rely on `ShiftedNabla` may compute incomplete tensor expressions.

---

## Fix

Use the shifted operator application instead of the plain gradient:

```
outer(f, s::ShiftedNabla) = transpose(s(f))
```

This ensures that the shift term `v ⊗ f` is included and restores the expected identity:

```
outer(f, s) == transpose(outer(s, f))
```

---

## Summary of changes

* Updated `outer(f, s::ShiftedNabla)` in
  `src/Fields/DiffOperators.jl`
* Replaced `transpose(gradient(f))` with `transpose(s(f))`
* Ensures the shift contribution `v ⊗ f` is correctly included
* Restores consistency between left and right operator forms

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `MappedGrid` now returns its internal geometric map. Since PR[#1226](https://github.com/gridap/Gridap.jl/pull/1226).
 - Fixed bug in skeleton transient fields. Since PR[1229](https://github.com/gridap/Gridap.jl/pull/1229).
 - Fixed bug in `FineToCoarseField` internal indexing. Since PR[1224](https://github.com/gridap/Gridap.jl/pull/1224).
+- Fixed `outer(f, s::ShiftedNabla)` silently ignoring the shift contribution.
 
 ## [0.19.8] - 2026-02-19
 
--- a/test/FieldsTests/DiffOperatorsTests.jl
+++ b/test/FieldsTests/DiffOperatorsTests.jl
@@ -54,6 +54,9 @@ g(x) = 2*x[2]
 @test (∇+p)⊗f != nothing
 @test f⊗(∇+p) != nothing
 
+s = ∇ - v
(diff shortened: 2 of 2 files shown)


Ady0333:
Hello @JordiManyer and @oriolcg , 
Please review this pr once when you have time. Happy to make any changes if needed!
Also i have raised an issue regarding this pr. Would like to have discussion if needed.

JordiManyer:
@Ady0333 Add a test that would otherwise break, edit NEWS.md

Ady0333:
@JordiManyer 
Thanks for the review!

I've added a regression test verifying that `outer(f, s)` matches `transpose(outer(s, f))` for `ShiftedNabla`, which fails without the fix. I also updated `NEWS.md` under the *Fixed* section.

Please let me know if any further changes are needed.

Issue #1245, opened 2026-03-08
Bug: Incorrect `else p == HEX` branch in EdgeBasedRefinement


Ady0333:
### **Description**

In `src/Adaptivity/EdgeBasedRefinement.jl`, the control flow for barycentric refinement includes the following branch:

if Dc == 2
  faces_list = (ref_nodes,Int32[],cells_to_refine)
elseif p == TET
  faces_list = (ref_nodes,Int32[],Int32[],cells_to_refine)
else p == HEX
  faces_to_refine = 1:num_faces(topo,2)
  faces_list = (ref_nodes,Int32[],faces_to_refine,cells_to_refine)
end

However, in Julia `else p == HEX` is parsed as an unconditional `else` block where the expression `p == HEX` is evaluated but ignored.

Therefore this branch effectively behaves as:

else
  p == HEX
  faces_to_refine = 1:num_faces(topo,2)
  faces_list = (ref_nodes,Int32[],faces_to_refine,cells_to_refine)
end

This means the condition `p == HEX` is never actually checked.

As a result, the HEX-specific code path executes for any polytope that is not handled by the previous conditions (`Dc == 2` or `p == TET`).

---

### **Expected behavior**

The intention appears to be to explicitly check for the HEX polytope. The branch should likely be written as:

elseif p == HEX
  faces_to_refine = 1:num_faces(topo,2)
  faces_list = (ref_nodes,Int32[],faces_to_refine,cells_to_refine)

This ensures that the HEX-specific logic is only executed when `p == HEX`.

Otherwise, a final fallback branch could be added to handle unsupported polytopes.

Ady0333:
@JordiManyer and @oriolcg, 
Please review this issue once when you have time, so that i can start working on this issue.

JordiManyer:
@Ady0333 Thanks for the issue. Sure, do a PR.

PS: If you have stuff like this, can you just do a PR and not do the issue? It just clutters everyone's feed.

Ady0333:
@JordiManyer 
Thanks for the feedback! I'll open a PR with the fix.

Also noted that for small fixes like this I'll submit a PR directly next time instead of opening an issue.

PR #1246, opened 2026-03-09
Fix get_tangent_vector and barycentric refinement HEX control flow

Ady0333:
This PR fixes the issue reported in #1245 and includes a small additional fix discovered while reviewing related code.

1. EdgeBasedRefinement control flow

`src/Adaptivity/EdgeBasedRefinement.jl:516`

The code previously used:

  else p == HEX

In Julia this is parsed as an unconditional `else` block where the expression `p == HEX` is evaluated but ignored. As a result, the HEX-specific branch executed for any polytope not handled by the previous conditions.

This PR replaces it with:

  elseif p == HEX

so that the condition is actually checked.

2. Copy-paste bug in `get_tangent_vector`

`src/CellData/CellFields.jl:590-591`

`get_tangent_vector` internally called `get_normal_vector` on each skeleton side instead of `get_tangent_vector`. This appears to be a copy-paste mistake from the overload above.

Both functions currently produce identical results, so behavior was unchanged, but the call should match the function’s intended semantics.

---

### **Impact**

- Fixes incorrect control flow in barycentric refinement logic.
- Prevents the HEX branch from being executed unintentionally for other polytopes.
- Corrects the internal call in `get_tangent_vector` so that it matches the function’s semantics.

---

### **Tests and NEWS**

- Added regression tests covering the refinement control flow and the tangent vector call.
- Updated `NEWS.md` under the *Unreleased → Fixed* section.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed bug in `FineToCoarseField` internal indexing. Since PR[1224](https://github.com/gridap/Gridap.jl/pull/1224).
 - Fixed `outer(f, s::ShiftedNabla)` silently ignoring the shift contribution.
 - Fixed incorrect `else p == HEX` control flow in `EdgeBasedRefinement.jl`. Since PR[1246] (https://github.com/gridap/Gridap.jl/pull/1246).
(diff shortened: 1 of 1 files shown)


Ady0333:
@JordiManyer 
Please review this pr once. I have also added a test and updated new.md.

Let me know if any changes are required, happy to make them !!

Issue #1231, opened 2026-02-25
compute_field_offsets applies forward permutation instead of inverse in BlockMultiFieldStyle


anusha19murthy:
## **Describe the bug**

`compute_field_offsets` in `MultiFieldFESpaces.jl` applies the forward permutation when recovering per-field offsets for `BlockMultiFieldStyle`.

During construction of a `MultiFieldFESpace` with a non-identity permutation `P`, block offsets are first computed following the block ordering induced by the permutation. However, the final offsets returned by `compute_field_offsets` must correspond to the original field ordering.

The current implementation uses:

```julia
offsets = map(p -> block_offsets[p], P)
```

which applies the forward permutation to block_offsets.

Since block_offsets already follows the permuted block ordering, recovering offsets per field requires applying the inverse permutation. Using the forward permutation incorrectly reorders offsets between fields.

As a consequence, offsets associated with one field may be assigned to another field.

**Impact**

Field offsets are later used by _restrict_to_field to extract each field’s DOF slice from the global block vector.

When offsets are incorrect:

fields may read DOF ranges belonging to other fields

DOF slices can overlap or become swapped

assembly and linear/nonlinear solves complete without errors

resulting solutions silently contain swapped or corrupted field values

This produces incorrect numerical results in multifield simulations using BlockMultiFieldStyle with explicit permutations.

The failure is silent and particularly difficult to detect since solver convergence is unaffected.

**Affected file**
src/MultiField/MultiFieldFESpaces.jl

**Possible fix**

Recover offsets using the inverse permutation instead of the forward permutation.

**Replace:**

offsets = map(p -> block_offsets[p], P)

with:

Pinv = invperm(P)
offsets = map(p -> block_offsets[p], Pinv)

This ensures offsets correspond to the correct field ordering and prevents incorrect DOF extraction across fields.

PR #1247, opened 2026-03-09
Fix/block multifield permutation

Ady0333:
This PR fixes the issue reported in #1231 related to incorrect permutation handling in `BlockMultiFieldStyle`.

## **Problem**

In `src/MultiField/MultiFieldFESpaces.jl`, the method `compute_field_offsets` computed offsets using the permutation `P` directly:

```
offsets = map(p -> block_offsets[p], P)
```

However, `block_offsets` is indexed by the **permuted position**, while the returned `offsets` vector must be indexed by the **original field index**.
Using `P` directly applies the permutation in the wrong direction.

As a result, when `BlockMultiFieldStyle` is used with non-trivial permutations, the computed offsets can be incorrect. This propagates to `get_cell_dof_ids`, leading to incorrect DOF placement during block assembly.

---

## **Fix**

Use the inverse permutation when mapping offsets:

```
offsets = map(p -> block_offsets[p], invperm(P))
```

This ensures each field receives the correct offset corresponding to its position in the permuted ordering.

Additionally, in `get_cell_dof_ids`, the previous implementation assumed the first field always had zero offset:

```
if i == 1
```

This assumption is not valid when fields are permuted. The condition has been updated to:

```
if offsets[i] == 0
```

so that the logic works correctly for any permutation.

---

## **Tests**

Added regression tests covering different `BlockMultiFieldStyle` permutations to ensure:

* `compute_field_offsets` returns correct offsets
* block matrix assembly produces correct block sizes

These tests would fail with the previous implementation and pass with the corrected one.

## **Files Modified**

* `src/MultiField/MultiFieldFESpaces.jl`

  * Fix permutation handling in `compute_field_offsets`
  * Update offset check in `get_cell_dof_ids`

* `test/MultiFieldTests/MultiFieldFESpacesTests.jl`

  * Add regression tests for permuted `BlockMultiFieldStyle`

* `NEWS.md`

  * Added changelog entry under **Unreleased → Fixed**

### DIFF
--- a/test/MultiFieldTests/MultiFieldFESpacesTests.jl
+++ b/test/MultiFieldTests/MultiFieldFESpacesTests.jl
@@ -133,9 +133,9 @@ X3 = MultiFieldFESpace([V1,V2,V3];style=BlockMultiFieldStyle(2,(1,2),(1,3,2)))
 
 ndofs = map(num_free_dofs,(V1,V2,V3))
 
-@test MultiField.compute_field_offsets(X1) == [0, ndofs[1], 0]
-@test MultiField.compute_field_offsets(X2) == [ndofs[3], 0, 0]
-@test MultiField.compute_field_offsets(X3) == [0, ndofs[3], 0]
+@test MultiField.compute_field_offsets(X1) == (0, ndofs[1], 0)
+@test MultiField.compute_field_offsets(X2) == (ndofs[3], 0, 0)
+@test MultiField.compute_field_offsets(X3) == (0, ndofs[3], 0)
 
 A1 = assemble_matrix(a,X1,X1)
 A2 = assemble_matrix(a,X2,X2)


Ady0333:
@JordiManyer, i have resubmitted the pr. Please go through it once.

JordiManyer:
@Ady0333 the fix for these error messages seems quite straighforward no? Any issue fixing it?

Ady0333:
> @Ady0333 the fix for these error messages seems quite straighforward no? Any issue fixing it?

Sorry for the delay @JordiManyer. Actually I was figuring it out and was having some trouble fixing them, will try to solve it soon.

JordiManyer:
Trouble? It just doesnt like comparing tuples to arrays no?

Ady0333:
Thanks for the hint. I suspected the issue might be related to comparing tuples with arrays, but I was still trying to track down exactly where the mismatch hapens. I wanted to make sure I understood the problem and fixed it properly rather than changing things blindly.

Ady0333:
@JordiManyer All the checks are have passed.

PR #1251, opened 2026-03-12
CI Improvements

JordiManyer:
The basic idea is to try to improve CI runtimes. I have split tests into different jobs, like I do in GridapSolvers. An additional benefit to this is that one can now run specific tests quite easily using the provided environment variable.


Antoinemarteau:
Have you also tried to write a small bash script that helps running tests in parallel locally ?

PR #1252, opened 2026-03-13
Add sign flip for nedelec hexa facet dofs

amartinhuertas:


Antoinemarteau:
Does this only work if the 3D FESpace is defined on a model with HEX only, or not ? (we'll have to adapt this to moment-based-reffes very soon)

PR #1271, opened 2026-04-05
fix autodiff of functions requiring a VectorValue

Dougal-s:
Autodifferentiating functions that require a `VectorValue` as input causes an error because the autodiff routines pass in an `SVector` instead.

For example, running
```julia
begin
	x₀ = VectorValue(1.0, 2.0)
	f(x) = (x - x₀) ⋅ (x - x₀)
	∇(f)(VectorValue(1.0, 2.0) + x₀)
end
```
results in
```julia
ERROR: MethodError: no method matching -(::StaticArraysCore.SVector{2, ForwardDiff.Dual{ForwardDiff.Tag{…}, Float64, 2}}, ::VectorValue{2, Float64})
For element-wise subtraction, use broadcasting with dot syntax: array .- scalar
The function `-` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  -(::Missing, ::Number)
   @ Base missing.jl:123
  -(::typeof(gradient), ::VectorValue)
   @ Gridap ~/.julia/packages/Gridap/cvsif/src/Fields/DiffOperators.jl:167
  -(::V, ::V) where V<:Gridap.TensorValues.MultiValue
   @ Gridap ~/.julia/packages/Gridap/cvsif/src/TensorValues/Operations.jl:94
  ...

Stacktrace:
 [1] f(x::StaticArraysCore.SVector{2, ForwardDiff.Dual{ForwardDiff.Tag{typeof(f), Float64}, Float64, 2}})
   @ Main ./REPL[2]:3
```

After this patch, I get
```
VectorValue{2, Float64}(2.0, 4.0)
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed type inference for operations between MultiValue types. Since PR[1270](https://github.com/gridap/Gridap.jl/pull/1270).
+- Allow autodifferentiation of functions that require a VectorValue input. Since PR[1271](https://github.com/gridap/Gridap.jl/pull/1271).
 
 ## [0.20.3] - 2026-03-23
 
--- a/src/Fields/AutoDiff.jl
+++ b/src/Fields/AutoDiff.jl
@@ -52,17 +52,17 @@ function gradient(f::Function,x::Point)
 end
 
 function gradient(f::Function,x::Point,fx::Number)
-  VectorValue(ForwardDiff.gradient(f,get_array(x)))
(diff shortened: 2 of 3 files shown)


JordiManyer:
Hi @Dougal-s thank you for the contribution. I am trying to understand: Was this an issue before? Or did something fundamentally change? I don't remember this being an issue? I am not sure.

Dougal-s:
Most of the code I changed hasn't been touched in 5 years, so I think it has always been like this. Inner products and indexing work the same with SVectors as with VectorValues, so I only recently noticed.

JordiManyer:
Ok, looks ok to me. Can you add tests that actually stress this and log the changes in NEWS?

JordiManyer:
Also @Dougal-s can you pull from master before I merge this? Cheers

PR #1274, opened 2026-04-09
Fix simplex uniform refinement for RT and Nedelec spaces.

DimhamT:
Before this PR, uniform refinement could produce fine meshes whose cell connectivity was not sorted consistently: the node indices inside a cell were not in ascending order. This caused incorrect construction of Raviart-Thomas and Nedelec FE spaces on the refined mesh. This PR fixes the problem by enforcing ascending local node ordering in each fine cell generated by uniform refinement.
I would appreciate a review from the maintainers.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed model orientation preservation in `UniformRefinement`. Since PR[#1274](https://github.com/gridap/Gridap.jl/pull/1274).
+
 ## [0.20.4] - 2026-04-09
 
 ### Added
--- a/src/Adaptivity/Adaptivity.jl
+++ b/src/Adaptivity/Adaptivity.jl
@@ -7,6 +7,7 @@ using Test
 using FillArrays
 using LinearAlgebra
 using DataStructures
+using StaticArrays
 
 using Gridap.Helpers
 using Gridap.Algebra
--- a/src/Adaptivity/UniformRefinement.jl
+++ b/src/Adaptivity/UniformRefinement.jl
@@ -125,8 +125,7 @@ function _permuted_simplex_rrule(::Val{Dc}, rr::RefinementRule, node_rank) where
     get_cell_type(ref_grid),
     OrientationStyle(ref_grid),
   )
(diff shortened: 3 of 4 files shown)


JordiManyer:
Hi @DimhamT , thanks for the PR. I have not looked at the code, except the fact that you only touched the files from your last PR. I dont really know what this code does at low level, and I do not use it. I also dont have the time to understand what it does. My feedback back when kt was first produced was that most of it is probably redundant, and I dont know why we have two versions of it. So if you think the changes are good, I have no problem with merging. 

In any case: I have been using RT and ND with my refinement routines, and have had no issues with it. So I am not sure the problem is with the spaces themselves. My best guess is that your routines generated non-oriented meshes, which is not supported at the moment. Other than this, I have no idea. 

So yeah, I am ok with merging if you think changes are ok.

DimhamT:
Hi, @JordiManyer . Thanks for your comment. Let me briefly explain the reason for this PR.
In 2D, the previous implementation of `UniformRefinement` assumed that every coarse cell could share the same `RefinementRule`. However, when uniform refinement inserts new nodes on edges, keeping the node ordering in the fine cells consistent requires the edge numbering of the coarse cell to follow a consistent convention as well, since this affects the consistency of the local-to-global mapping in `UniformRefinement`. 
For example,
```julia
cm = simplexify(CartesianDiscreteModel((0,1,0,1),(2,2)))
topo = get_grid_topology(cm)
get_faces(topo, 2, 1)

# output
# [1, 2, 3]
# [3, 4, 5]
# [6, 4, 7]
# ...
```
the edge ordering is not increasing. This is not an issue for Lagrangian elements, since they can handle local reordering through `permutation > 1` in [ConformingFESpaces.jl](https://github.com/gridap/Gridap.jl/blob/b6d2b0c606f4e067903e97c427ffb7dc979a8fae/src/FESpaces/ConformingFESpaces.jl#L853).

This PR address this by constructing the corresponding `RefinementRule` from the local ordering and local permutation of each d-face of the coarse cell.

I also verified that the expected convergence orders for RT and ND are recovered under uniform refinement after these changes.

JordiManyer:
Yeah so basically you were not preserving the orientation of the model. 
Again, if you are happy with the changes, I am ok with them. Log the changes in NEWS.md.

DimhamT:
I have updated NEWS.md accordingly. Thank you for your feedback and for merging.

PR #1283, opened 2026-04-24
[Fix]: Check convergence in NLSolver before returning solution

### LINKED ISSUE #1282: [bug]: NLSolver returns wrong solutions when Newton doesn't converge (no error thrown)
I was testing different nonlinear solvers for my neural operator ROM project and ran into something weird with NLSolver.

When Newton fails to converge, it just returns whatever the last iterate was - no error, no warning. I only found out because I was comparing results.

## Quick test case

I made a simple nonlinear problem where I force Newton to fail:

```julia
# Problem where true solution is x = 2.0
# Set max iterations low so it won't converge
solver = NLSolver(show_trace=false, method=:newton, iterations=5)

uh, cache = solve!(uh, solver, op)
println("Solution: ", uh)  # Got x = 44.44
println("Converged? ", cache.result.f_converged)  # false
println("Residual: ", cache.result.residual_norm)  # 87788
# No error thrown!
```

The solution is way off (44 vs 2) but Gridap accepts it.

## Comparison with NewtonRaphsonSolver

When I run the same problem with `NewtonRaphsonSolver`, it throws an error (which is what I'd expect). So this seems specific to NLSolver.

## Why I think this is a problem

1. NLSolver is the default - `NonlinearFESolver()` uses it internally
2. For FE problems, I got a FEFunction that looked normal but had 34% worse error than the converged solution
3. In transient problems this could be really bad - accepting unconverged steps silently

## Where the bug is

Looking at the code in `src/Algebra/NLSolvers.jl` around line 150:

```julia
r = nlsolve(df,x;linsolve=linsolve!,kwargs...)
cache.result = r
copy_entries!(x,r.zero)  # copies result without checking r.f_converged
```

It just copies the result without checking if nlsolve actually converged.

Same issue in `src/ODEs/StageOperators.jl` line 193.

## Suggested fix

Add a check:

```julia
r = nlsolve(df,x;linsolve=linsolve!,kwargs...)
cache.result = r
if !converged(r)
    error("Newton didn't converge after $(r.iterations) iterations, residual = $(r.residual_norm)")
end
copy_entries!(x,r.zero)
```

Ady0333:
## Summary

Fixes #1282 

When `nlsolve()` fails to converge, NLSolver now checks the convergence status before returning the solution. Previously, it silently copied the unconverged last iterate.

---

## Changes

- `src/Algebra/NLSolvers.jl`: Add `converged()` check after `nlsolve()` call
- `src/ODEs/StageOperators.jl`: Same fix for ODE/transient context

---

## Behavior

**Before:** Unconverged Newton iterates silently returned as valid solutions
**After:** Throws error with iteration count and residual norm

---

## Testing

Tested with a nonlinear problem that fails to converge:
- Previously: returned x=44.44 (wrong) with no error
- Now: throws "Newton solver did not converge after 5 iterations. Final residual norm: 87788"

All existing Gridap tests pass.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- `NLSolver` now checks convergence status before returning solution. Previously, when `nlsolve()` failed to converge, the unconverged last iterate was silently returned. Now throws an informative error with iteration count and residual norm. Since PR[#1283](https://github.com/gridap/Gridap.jl/pull/1283). (@Ady0333)
+
 ## [0.20.4] - 2026-04-23
 
 ### Added
--- a/src/Algebra/NLSolvers.jl
+++ b/src/Algebra/NLSolvers.jl
@@ -149,6 +149,9 @@ function _nlsolve_with_updated_cache!(x,nls,op,cache)
   end
   r = nlsolve(df,x;linsolve=linsolve!,kwargs...)
   cache.result = r
+  if !converged(r)
+    error("Newton solver did not converge after $(r.iterations) iterations. Final residual norm: $(r.residual_norm)")
+  end
   copy_entries!(x,r.zero)
 end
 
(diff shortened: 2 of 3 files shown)


Ady0333:
Hello @JordiManyer and @oriolcg!!! Please review this pr and let me know if any changes are needed...

JordiManyer:
Thanks @Ady0333 for the contribution, I don't see any issue with the PR. Could you please update NEWS.md with the changes? Cheers!

Ady0333:
> Thanks @Ady0333 for the contribution, I don't see any issue with the PR. Could you please update NEWS.md with the changes? Cheers!

Thanks for the review...I have updated the NEWS.md file.

PR #1285, opened 2026-04-24
[Fix]: NewtonRaphsonSolver convergence check and error handling

### LINKED ISSUE #1284: [bug]: NewtonRaphsonSolver crashes with @unreachable when starting from converged solution
I was testing Newton solvers and found that NewtonRaphsonSolver crashes if you start it from an already-converged state (which can happen in transient problems when you reuse the previous time step).

## Test case

```julia
# Start Newton at the exact solution
x_exact = [1.0, 2.0]
solver = NewtonRaphsonSolver(LUSolver(), 1e-10, 10)

uh = copy(x_exact)
solve!(uh, solver, op, nothing)
# ERROR: This line of code cannot be reached (@unreachable)
```

The residual at x_exact is ~0.0, so the convergence criterion becomes:
```julia
m < 1e-10 * 0.0  # i.e., m < 0, which is impossible
```

Newton runs all 10 iterations and then hits `@unreachable` at line 67.

## Why this happens

In `src/Algebra/NLSolvers.jl` line 79-82:

```julia
function _check_convergence(nls,b)
    m0 = maximum(abs,b)
    return (false, m0)  # Always returns false!
end
```

The initial convergence check is hard-coded to `false`, so even if you start with zero residual, it never exits early.

Then in the loop, when `m0 ≈ 0`, the relative tolerance `m < nls.tol * m0` is unsatisfiable, so it iterates until max and crashes.

## When this happens in real code

- Transient simulations reusing previous step as initial guess
- Steady state reached (residual already ~0)
- Continuation/homotopy methods
- Any time Newton re-enters with a very small residual

## Simple fix

Line 81, change:
```julia
return (false, m0)
```
to:
```julia
return (m0 < nls.tol, m0)
```

And line 67, replace `@unreachable` with a proper error:
```julia
error("NewtonRaphsonSolver failed to converge in $(nls.max_nliters) iterations, residual = $m")
```

I can make a PR for this if it looks right.

Related: #1282  - found while testing Newton solver convergence behavior

Ady0333:
## Summary

Fixes #1284 

Two related fixes for NewtonRaphsonSolver:

1. **Initial convergence check** - Actually test if starting residual is below tolerance
2. **Error handling** - Replace `@unreachable` with informative error message

---

## Changes

In `src/Algebra/NLSolvers.jl`:

**Line 81** - Initial check:
```julia
# Before: return (false, m0)
# After:  return (m0 < nls.tol, m0)
```

**Line 67** - Non-convergence handling:
```julia
# Before: @unreachable
# After:  error("NewtonRaphsonSolver failed to converge in $(nls.max_nliters) iterations. Final residual: $m")
```

---

## Behavior

**Before:**
- Starting from converged state → crash with "@unreachable"
- Relative tolerance `m < tol*m0` unsatisfiable when `m0 ≈ 0`

**After:**
- Already-converged initial state → exits immediately
- Non-convergence → informative error with diagnostics

---

## Testing

Tested with Newton starting at exact solution (zero residual):
- Previously: crashed with "@unreachable"
- Now: exits cleanly on first convergence check

All existing Gridap tests pass.

---

## Related

Related to #1282  - part of broader Newton solver convergence testing

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- `NewtonRaphsonSolver` now correctly handles already-converged initial states and reports non-convergence properly. Previously crashed with `@unreachable` when starting residual was ~0. Now checks initial convergence and throws informative error instead of assertion failure. Since PR[#1285](https://github.com/gridap/Gridap.jl/pull/1285). (@Ady0333)
+
 ## [0.20.4] - 2026-04-23
 
 ### Added
--- a/src/Algebra/NLSolvers.jl
+++ b/src/Algebra/NLSolvers.jl
@@ -64,7 +64,7 @@ function _solve_nr!(x,A,b,dx,ns,nls,op)
     if isconv; return; end
 
     if nliter == nls.max_nliters
-      @unreachable
+      error("NewtonRaphsonSolver failed to converge in $(nls.max_nliters) iterations. Final residual: $(maximum(abs,b))")
     end
 
     # Assemble jacobian (fast in-place version)
@@ -78,7 +78,7 @@ end
 
 function _check_convergence(nls,b)
   m0 = maximum(abs,b)
-  return (false, m0)
+  return (m0 < nls.tol, m0)
 end
 
 function _check_convergence(nls,b,m0)


Ady0333:
Hello @JordiManyer , @Antoinemarteau and @oriolcg !!! Please review this pr and let me know if any changes are needed...

~Thanks

JordiManyer:
Thanks @Ady0333 for the contribution. Changes look good, please update NEWS.md so that I can merge. This PR could have been bundled with the other one, it's addressing the same kind of things. 

Also, kindly stop tagging everyone when you open a PR. People have their own lives and our inboxes are already quite full. When you leave a PR, someone will eventually look into it. Unless your PR has been sitting without response for over a week, it is really not necessary.

Ady0333:
> Thanks @Ady0333 for the contribution. Changes look good, please update NEWS.md so that I can merge. This PR could have been bundled with the other one, it's addressing the same kind of things.
> 
> Also, kindly stop tagging everyone when you open a PR. People have their own lives and our inboxes are already quite full. When you leave a PR, someone will eventually look into it. Unless your PR has been sitting without response for over a week, it is really not necessary.

Thanks for the review. I have updated the NEWS.md and will make sure not to tag anyone...apologies for that!!!

PR #1290, opened 2026-04-29
Bugfix: testvalue for LinearAlgera.Diagonal

JordiManyer:


JordiManyer:
@Antoinemarteau we have an issue... `LinearAlgebra.Diagonal` needs to be a square matrix, but the current `return_value` for `LinearCombinationVector` wants to generate a 0xn_basis, so we actually change the type of what we return. I added tests that currently fail. 

This is quite horrible, and I do not see a way out other than removing the diagonals in favor of plain matrices, or implementing our own flavor of  diagonal coefficients that allows for non-square matrices. What do you think?

Antoinemarteau:
Or, change testvalue of `CompWiseTensorPolyBasis` (and all other bases) to return a zero length basis, like `CartProdPolyBasis` does ?
The only polynomial basis for which it is not possible to make it empty is `NedelecPolyBasisOnSimplex`, but it is quite unused ...

Antoinemarteau:
Alternatively, testvalue of LinearCombinationFieldVector could simply be consistent with the length of the basis returned by testvalue(F)?
``` values = zeros(eltype(V), length(fields), 0)```
becomes
``` values = Diagonal(zeros(eltype(V), length(fields), length(fields)))```
if V <: Diagonal ?

File: src/ReferenceFEs/GeometricDecompositions.jl [hunk 1]
```diff
@@ -197,7 +197,7 @@ function apply_face_signflip(
 
   facet_range = get_dimrange(p,D-1)
   face_own_funs = get_face_own_funs(b,p,conf)
-  sign_flip = MVector(tfill(1, Val(length(b)))...)
+  sign_flip = fill(1, length(b))
```
JordiManyer, line 200: @Antoinemarteau Not only you are using Diagonal, you were using Diagonal of weird stuff, like `MVectors` and `Ones`. The worse part is that you can't even use `similar` on those, because they return sregular arrays. Its a mess... is it really worth it the optimisation? If you want to keep it like this, could you have a look on how to make it viable? I guess this will only appear for linear combinations, but the amount of possibilities is quite bad right now.

Antoinemarteau: It's probably not worth it to use MVectors at evaluation time, and might not change anything if the evaluation of LinearCombination[...] uses @inbounds, as the size can be determined from the `field`...
The downside would be allocations when building the reffes (only an issue if many are build).

src/Fields/FieldArrays.jl
```diff
@@ -202,7 +204,7 @@ end
 
 function testvalue(::Type{LinearCombinationField{V,F}}) where {V,F}
   fields = testvalue(F)
-  values = testvalue(V)
+  values = zeros(eltype(V), length(fields), 0)
```
Antoinemarteau, line 207: This still looks weird to me, `V` is ignored so the returned testvalue isn't necessary of the same type as the argument, same with the FieldVector version (if V isn't Diagonal). Doesn't this break the API of testvalue ?

src/ReferenceFEs/GeometricDecompositions.jl
JordiManyer: Also, unless you have MVectors everywhere I dont think there is any loop unrolling

PR #1291, opened 2026-04-29
Strengthen/bug-fix compute_facet_owners

amartinhuertas:


src/FESpaces/Pullbacks.jl
```diff
@@ -188,31 +188,20 @@ function evaluate!(cache,k::NormalSignMap,reffe,facet_own_dofs,cell)
   return Diagonal(dof_sign)
 end
 
-function compute_facet_owners(model::DiscreteModel{Dc}, local_to_global::F=identity) where {Dc,F}
+function compute_facet_owners(model::DiscreteModel{Dc}, select_nbor=first) where {Dc}
```
amartinhuertas: I would not use `first` by default, but `maximum` better.

JordiManyer: Sure, I guess it would make it a little more consistent.

amartinhuertas: ok, done

PR #1303, opened 2026-05-19
Fixed type instability in array_cache for LazyArray

JordiManyer:
I've been tracking this type instability for some time: with the current implementation, we have

```julia
function array_cache(dict::Dict,a::LazyArray)
  cache = _get_cache(dict,a)
  if cache === nothing
    _cache = _array_cache!(dict,a)
    dict[objectid(a)] = (a,_cache)
  else
    _cache = cache
  end
  _cache
end
```

The issue here is that, when we hit the cache and have `!isnothing(cache)`, the only thing the compiler can see is that the cache comes from a `Dict{UInt,Any}` and thus sets the return type to `Any`. 

This is HORRIBLE, since it generates a type instability that hinders the evaluation of every single subsequent call to `getindex!`, since that `Any` typing is baked into the cache. 

After exploring several solutions, I think we should just pay the price of re-computing the caches. We keep the memoization and the memory savings (the recomputed cache gets thrown away), but we are now able to tell the compiler what type it should expect. The final version is: 

```julia
function array_cache(dict::Dict,a::LazyArray)
  cache = _array_cache!(dict,a)
  existing = _get_cache(dict,a)
  T = typeof(cache)
  if isnothing(existing)
    dict[objectid(a)] = (a,cache)
    return cache::T
  else
    dict[objectid(a)] = (a,existing)
    return existing::T
  end
end
```

This is quite delicate, so I would appreciate a review @amartinhuertas @Antoinemarteau

### DIFF
--- a/src/Arrays/LazyArrays.jl
+++ b/src/Arrays/LazyArrays.jl
@@ -128,23 +128,16 @@ IndexStyle(::Type{<:LazyArray{G,T,1} where {G,T}}) = IndexLinear()
 uses_hash(::Type{<:LazyArray}) = Val{true}()
 
 function same_branch(a,b)
-  a === b
+  return a === b
 end
 
 function same_branch(a::Fill,b::Fill)
   typeof(a) != typeof(b) && return false
   size(a) != size(b) && return false
-  a.value == b.value
+  return a.value == b.value
 end
 
-function all_same_branch(a::Tuple,b::Tuple)
-  for i in 1:length(a)
-    if same_branch(a[i],b[i]) == false
-      return false
-    end
-  end
-  true
-end
+all_same_branch(a::Tuple,b::Tuple) = all(map(same_branch,a,b))
 
 function same_branch(a::LazyArray,b::LazyArray)
   typeof(a) != typeof(b) && return false
@@ -155,26 +148,32 @@ end
 function _get_cache(dict,a)
   id = objectid(a)
   if haskey(dict,id)
-    o,c = dict[id]
+    o, c = dict[id]
     return c
   end
-  for item in dict
-    if same_branch(a,second(item)[1])
-      return second(item)[2]
+  for item in values(dict)
+    o, c = item
+    if same_branch(a,o)
+      return c
     end
   end
   return nothing
 end
 
+# We recompute the cache even if we find an existing one, 
+# because otherwise we CANNOT infer the type and we get 
+# a type instability for every subsequent getindex! call.
 function array_cache(dict::Dict,a::LazyArray)
-  cache = _get_cache(dict,a)
-  if cache === nothing
-    _cache = _array_cache!(dict,a)
-    dict[objectid(a)] = (a,_cache)
+  cache = _array_cache!(dict,a)
+  existing = _get_cache(dict,a)
+  T = typeof(cache)
+  if isnothing(existing)
+    dict[objectid(a)] = (a,cache)
+    return cache::T
   else
-    _cache = cache
+    dict[objectid(a)] = (a,existing)
+    return existing::T
   end
-  _cache
 end
 
 # to memoize last access of a lazy array
@@ -190,16 +189,13 @@ end
 
 function _array_cache!(dict::Dict,a::LazyArray)
   @check begin
-    @notimplementedif ! all(map(isconcretetype, map(eltype, a.args)))
-    if !(eltype(a.maps) <: Function)
-      @notimplementedif !isconcretetype(eltype(a.maps))
-    end
-    true
+    all(map(isconcretetype, map(eltype, a.args))) &&
+      (eltype(a.maps) <: Function || isconcretetype(eltype(a.maps)))
   end
   gi = testitem(a.maps)
   fi = map(testitem,a.args)
   cg = array_cache(dict,a.maps)
-  cf = map(fi->array_cache(dict,fi),a.args)
+  cf = map(Base.Fix1(array_cache, dict),a.args)
   cgi = return_cache(gi, fi...)
   index = -1
   #item = evaluate!(cgi,gi,testargs(gi,fi...)...)
--- a/test/ArraysTests/LazyArraysTests.jl
+++ b/test/ArraysTests/LazyArraysTests.jl
@@ -326,4 +326,22 @@ c = a + b
 @test all(lazy_map(test_map,c,k,a,b))
 @test all(lazy_map((x...)->test_map(x...,cmp=≈),c,k,a,b))
 
+# array_cache type stability: compiler must infer a concrete type (not Any)
+# on both the miss path and the hit path (DAG — same node as multiple args).
+let
+  x = [1.0, 2.0, 3.0]
+  y = lazy_map(a -> a .+ 1.0, x)
+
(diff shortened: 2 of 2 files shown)


Antoinemarteau:
My though about this:

- This PR trades a dynamic dispatch on `getindex!` with a type assert in `array_cache` (that will always be dynamic/verified at runtime, since `typeof(existing)` is inferred as `Any`), I genuinely don't know which is cheaper, did you benchmark it ? I suspect they have similar costs.
- if there is a function barrier between getting the cache and using `getindex!` on it, the type stability of `array_cache` is possibly not important, you get one dynamic dispatch for the barrier, and inside the type of the cache is known. 
 
But since there are tons of instances of not having a function barrier in the code, and with several or lots of `getindex!` call for a single `array_cache`, I think that this PR is a good idea.

Antoinemarteau:
Hold on a second, there's zero instance of calling `array_cache(dict,a)` in the whole code outside `Arrays` module, are you sure that this memoization is ever used? It seems to me that all `array_cache(a)` calls (on the "outermost" LazyArray) creates a new dict, put their cache in there, and it's never re-used, or am I missing something ?

Other avenue: since `array_cache(a)` specializes on `typeof(a)`, if we assume that `typeof(cache)` is unique for each `typeof(a)`, then we can change this generic 
```
_default_array_cache(a,s::Val{true}) = array_cache(Dict{UInt,Any}(),a)
```
to actually create a first `cache`, and put `typeof(cache)` instead of `Any` for that dict, to recover type stability of geting in the dict.

Antoinemarteau:
More info after some discussion with @JordiManyer  and some work:
- the point of the dict is to keep only one cache for each repeated identical operation in the operation tree of the parent array, especially to mutualize the memoized `IndexItemPair`value in the lazy arrays caches.
- the goal here is to make the concrete `typeof(cache)` inferable at compile time (to avoid dynamic dispatches in `getindex!` in the context of the caller of `array_cache`), it is not only about making `array_cache` type stable.
- the current solution above achieves this, but at the price of re-computing all the caches that could already be stored. I benchmarked `array_cache(dict, lazyarray)` for a large domain-contribution `LazyArray` from the PoissonDGTests, and get
-- `2.138 μs (8 allocations: 5.36 KiB)`, and inferred type `Any` for the existing code
-- `44.919 μs (1688 allocations: 153.92 KiB)`, and correctly inferred `typeof(cache)` from the solution above
-- `3.720 μs (10 allocations: 6.98 KiB)`, and correctly inferred `typeof(cache)`, with the very hacky solution
-- `1.449 μs (1 allocation: 576 bytes)`, and correctly inferred `typeof(cache)`, with the very hacky solution and removing `same_branch`.
-- `24.496 μs (1445 allocations: 118.17 KiB)`, and correctly infered `typeof(cache)`, with the above solution and removing `same_branch`.
(these benchmarks were made with `@btime`, so it's the best case where all caches are already in the `dict`)

Hacky solution: `_array_cache_type` and using aggressive `@assume_effect` to constant fold ( it's a `Core.Compiler.return_type` taking argument value instead of their type, on steroid)
```julia 

function _array_cache_type(dict,a::LazyArray)
  cache = _array_cache!(dict, a)
  return typeof(cache)
end

function array_cache(dict::Dict,a::LazyArray)
  T = begin Base.@assume_effects :foldable :nothrow _array_cache_type(dict,a) end
  existing = _get_cache(dict,a)
  if isnothing(existing)
    cache = _array_cache!(dict,a)
    dict[objectid(a)] = (a,cache)
    return cache::T
  else
    existing
    dict[objectid(a)] = (a,existing)
    return existing::T
  end
end
```

We also discussed the opportunity to get rid of the `same_branch` logic which seems to be not-alway-good micro-optimization that complexifies the code.  It was added in #701, but unfortunately there is no benchmark added there to detect performance regression of ditching that. If we remove `same_branch`, we can replace the `Dict` with an `IdDict` and elegantly rewrite the function as:
```julia
function array_cache(dict::IdDict, a::LazyArray)
  T = begin Base.@assume_effects :foldable :nothrow _array_cache_type(dict,a) end
  # T = _array_cache_type(dict,a) # this actually allocates all the children's caches..
  # exchange the two lines above to get the "existing solution with removed same_branch benchmark"

  cache = get!(dict, a) do
    _array_cache!(dict, a)
  end

  return cache::T
end
```
The `Arrays` and `FESpaces` tests passed after I removed `same_branch`.

Antoinemarteau:
Ok, I think I found an ideal solution: caches will only be computed once, and the only dynamic thing happening is the type assertion `cache::T` at the end of `array_cache`. Code is as follows:
```julia
function array_cache(dict::IdDict, a::LazyArray)
  T = _allocate_cache_and_compute_its_type(dict,a)
  cache = get(dict, a) do
    @unreachable "The cache should be stored already."
  end
  return cache::T
end

function _allocate_cache_and_compute_its_type(dict, a::LazyArray)
  cache = get!(dict, a) do
    _array_cache!(dict, a)
  end
  return typeof(cache)
end
```
This code assumes that `same_branch` was removed, and that the `Dict` has been replaced with an `IdDict`, but the trick can be adapted even if we keep `same_branch`.

Some benchmarks using `@btime` (caches are in the dict):
- If the cache is not put in the dict while computing `T` :` 24.446 μs (1445 allocations: 118.17 KiB)` (=== to solution at the top)
- If the cache is stored in the dict while computing `T` :  `2.863 μs (2 allocations: 1.12 KiB)` (this solution)
- hacky solution with `Base.@infer_effects` :  `1.447 μs (1 allocation: 576 bytes)`

Benchmark using `@time` (caches are not in the dict) 
- previous solution: `200-300 μs (6.59 k allocations: 533.250 KiB)`
- new solution:        `150-170 μs (1.49 k allocations: 126.172 KiB) `
- super hacky:          `110-130 μs (1.45 k allocations: 121.281 KiB)`

Antoinemarteau:
@JordiManyer I think I should implement the last solution on this branch, do we agree that we can remove `same_branch` ?

JordiManyer:
> ```julia
> function array_cache(dict::IdDict, a::LazyArray)
>   T = _allocate_cache_and_compute_its_type(dict,a)
>   cache = get(dict, a) do
>     @unreachable "The cache should be stored already."
>   end
>   return cache::T
> end
> 
> function _allocate_cache_and_compute_its_type(dict, a::LazyArray)
>   cache = get!(dict, a) do
>     _array_cache!(dict, a)
>   end
>   return typeof(cache)
> end
> ```

Oh I like this, I like it very much indeed. Let me look at this tomorrow to test myself. 

Concerning same_branch, let me also have a look. In the PR you mention, the conclusion seems to be that it matters. Maybe @ericneiva remembers in which cases it mattered?

ericneiva:
Hello, @JordiManyer and @Antoinemarteau,

> Maybe @ericneiva remembers in which cases it mattered?

I barely remember and I feel bad about this, but I do know the short answer is _it matters in weak forms with a large number of terms_.

IIRC, here is the example that motivated this optimization.

[transient_navierStokes.txt](https://github.com/user-attachments/files/28152196/transient_navierStokes.txt)

It's an implementation of an Embedded ALE formulation of the time-dependent Navier-Stokes equations.

Hope this gives you some hints.

PR #1306, opened 2026-05-24
Dof permutations for moment-based  reffes

JordiManyer:
The strategy: 

- New API `get_basis_permutations(poly::Polytope, basis)` that returns the permutations of the moment bases for a certain face polytope. 
- Then `moment_face_own_dofs_permutations` takes the moments and computes the complete set of permutations by combining the permutations of the moment bases on each face. 

Details to figure out: 

- Not all bases are invariant through permutations. In particular, some of the bases used for the moments are not. This raises two questions: 
  - Can we change the bases we use for RT/ND/... to use exclusively bases that are invariant under permutations? 
  - Can we modify how we construct some of the current bases so that they are invariant under permutations?
  - For the bases which are NOT invariant under permutations, what do we return as the default? Right now I am returning the identity perm, but that will yield wrong results silently (which is bad). 
- Currently, the implementation only works for `change_dof=false`. In the case where `change_dof=true`, the final dofs do not correspond to the moments. I am not sure how to deal with this... Should the dof ownership be attached to the prebasis instead? No idea... 
- Currently I have place the permutations within the MomentBasedDofBasis. It was the simplest way to test things. I am open to reconsider its placement. 

I have added a MWE for raviart-thomas on meshed with flipped edges. Right now, we have working: RT on simplices (TRI/TET). RT on QUAD/HEX could work if we changed the default basis to use Bernstein. ND currently depends on bases which are not permutation invariant. 

I have also added a small driver that helps inspecting the bases permutations for each combination of polytope and basis. @Antoinemarteau


Antoinemarteau:
Update on this:
- the `get_basis_permutations(poly::Polytope, basis)` API should be implemented for the elemental bases too (not only face moment basis) in order to solve the issue for `change_dof=true`
- It is not possible to have "permutation" of all the vector-valued moment bases on simplicies, i.e. the moment bases of 3D Nédélec on TET of both kinds.
-  It should be possible for n-cubes basis, except D>=3 Serendipity of order >= 6 (you get a nontrivial barycentric basis on a square face..)
- For bases that don't have a permutation, we need a face-wise change of basis matrix between the face basis. It can be stored in the ReferenceFE.
- I would place the permutations in GenericRefFE, this would make more sense for `change_dof=true` (where perms are determined by shapefuns)

The implementation of these face wise changes is not that trivial, since currently the permutation info (face_own_dofs_permutations) is used in CellConformity "only", this only works because there are 1 - 1 relations (optionally signed) between dofs. 
Once we need to apply (face-wise) changes of basis, which relate several dofs with each other, this has to be done in FESpaces/Pullback.jl, in `compute_cell_bases_changes`, in a consistent manner with how the CellConformity defines the global dofs.

Lastly, the permutation test driver needs to be updated for vector valued bases, since the correct Piola map needs to be applied when mapping the polynomials to the permuted face.

Antoinemarteau:
Additional remarks:
- since the validity of "permutations" of vector valued polynomial depend on the Piola maps, I expect that the `get_basis_permutations` needs to have an additional `conf::Conformity` argument in order to be implemented for `CompWiseTensorPolyBasis` and `BarycentricyP(m)ΛBasis`, similarly to the geometric decomposition APIs.
- Since we'll have to implement those face-wise changes of bases, we can implement a general default numeric version that can glue any bases, as long as their trace spaces on a given face are the same.

PR #1316, opened 2026-06-11
Bugfix empty trians with inverse maps

ericneiva:
Hello @JordiManyer,

I've implemented this to integrate weak forms on empty triangulations with inverse maps.

I've been suggested to also dispatch V on `Diagonal` for the testvalue of LinearCombinationField. Let me know if we should not have it.

Thanks a lot!


JordiManyer:
Looks good! Thanks @ericneiva !

PR #1211, opened 2026-01-20
Proposal: added `normalize` function

miguelmaso:
Hi,

Thank you again for developing and maintaining this package, it's very handy for developing new formulations.

I'd like to make the `normalize` function from `LinearAlgebra` available for `VectorValues`. This PR shows a possible implementation, another option would be to rely on the `StaticArrays` library.

Does it make sense to you? Thanks in advance!

PR #1317, opened 2026-06-16
Consistent Linear-Algebra exports: normalize and eigen

miguelmaso:
In #1157 and #1211 there were added the corresponding dispatches for `LinearAlgebra.eigen` and `LinearAlgebra.normalize` with `TensorValue`. These functions are currently exported in `TensorValues` module, but not in the `Gridap` namespace.

This PR makes the exports consistent between
https://github.com/gridap/Gridap.jl/blob/eabb7e568a19a822f2fbbef82e4e7833bd2f837b/src/TensorValues/TensorValues.jl#L108
and
https://github.com/gridap/Gridap.jl/blob/eabb7e568a19a822f2fbbef82e4e7833bd2f837b/src/Exports.jl#L10

### DIFF
--- a/src/Exports.jl
+++ b/src/Exports.jl
@@ -6,8 +6,8 @@ macro publish(mod,name)
 end
 
 # Reexport from LinearAlgebra (just for convenience)
-using LinearAlgebra:  det, inv, tr, cross, dot, norm, ×, ⋅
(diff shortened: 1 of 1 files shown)


miguelmaso:
Just for convenience, let me know if it makes sense

JordiManyer:
Sure, why not. Thanks @miguelmaso !

PR #1325, opened 2026-07-11
Add AD support for complex-valued DOFs

zjwegert:
This PR adds basic autodiff support for problems with complex-valued DOFs. Currently, this is not supported natively because ForwardDiff does not fully support complex-valued functions.

The approach here is to split the computation into real and imaginary parts, and since we're assuming that the functional is differentiable, we can compute the derivative by only perturbing the real parts of the cell dofs. I thought the best approach would be to dispatch on `T=eltype(get_vector_type(V))` and add new methods `autodiff_array_gradient(::Type{<:Complex},...)` and `autodiff_array_gradient(::Type{<:Complex},...)` that split the computation as above.

There are three cases that we consider:
1. Holomorphic maps from `ℂ` → `ℂ`. The standard approach in AD systems is to compute the conjugate gradient `f'(z)ᴴ = ∂ᵣu + i ∂ₛu` so that `dF(z)(v) = <f'(z)ᴴ, v> = f'(z)v`.
2. Non-holomorphic maps from `ℂ → ℝ`. Using CR-Calculus we have that the gradient is given by `∇z(f) = 2∂f/∂z* =  ∂ᵣu + i ∂ₛu` and the directional derivative is given by `dF(z)(v)=Re{∇z(f),v}`.
3. (NOT IMPLEMENTED) General non-holomorphic maps from `ℂ → ℂ`: requires splitting into derivative in `z` and derivative in `z*`. This is not implemented yet. To do this one properly, we should probably implement a new kind of FEOperator called `NonholomorphicFEOperator` that appropriately builds the Jacobian's in `z` and `z*` and implements a basic Newton method.

Because (1) and (2) are in the same form, we reuse the code below for both cases.

@JordiManyer, If we're happy with this PR, I'll make the appropriate changes to GridapDistributed as well.


src/Arrays/Autodiff.jl
```diff
@@ -61,10 +61,59 @@ function autodiff_array_jacobian(a,i_to_x,j_to_i;tag=default_tag(ForwardDiff.jac
...
+  s = lazy_map(Broadcasting(imag),i_to_x)
+  g = r -> f(lazy_map((r,s) -> r + im * s, r, s))
+  f₁ = r -> lazy_map(Broadcasting(real),g(r))
+  f₂ = r -> lazy_map(Broadcasting(imag),g(r))
+  f₁,f₂
+end
+
+"""
+"""
+function autodiff_array_gradient(::Type{<:Complex},a,i_to_x;tag=default_tag(ForwardDiff.gradient,a))
```
JordiManyer, line 93: Can we maybe reuse this structure? Like define a `autodiff_array_complex_work` that does the splitting + merging? It would reuse most of the code.

src/Arrays/Autodiff.jl
```diff
@@ -61,10 +61,59 @@ function autodiff_array_jacobian(a,i_to_x,j_to_i;tag=default_tag(ForwardDiff.jac
...
+  autodiff_array_jacobian(V,agrad,i_to_x,j_to_i;tag=agrad_tag)
+end
+
+# For complex case we compute the derivative of f as f'(x) = ∂ᵣf₁ + i*∂ᵣf₂ via Cauchy-Riemann
+# equations where f(x) = f₁(r,s) + i*f₂(r,s) and x = r + i*s. I.e., in this case we perturb
+# only the real part and keep the imaginary part fixed.
+function _change_argument_real(f,i_to_x)
+  s = lazy_map(Broadcasting(imag),i_to_x)
+  g = r -> f(lazy_map((r,s) -> r + im * s, r, s))
+  f₁ = r -> lazy_map(Broadcasting(real),g(r))
```
JordiManyer: Any reason we are using anonymous functions instead of declaring `g(r)` inside the function? I think there is difference compiler-wise.

src/FESpaces/FEAutodiff.jl
```diff
@@ -200,4 +206,48 @@ function Arrays.autodiff_array_jacobian(a, i_to_x, j_to_i::SkeletonPair; tag=def
   is_single_field = eltype(eltype(j_to_result_plus)) <: AbstractArray
   k = is_single_field ? Arrays.MergeBlockMap((2,2),I) : Arrays.BlockBroadcasting(Arrays.MergeBlockMap((2,2),I))
   lazy_map(k,j_to_result_plus,j_to_result_minus)
+end
+
+# Skeleton + Complex
+# Evaluate r_dual + im*s, then extract the complex gradient/jacobian
+function Arrays.autodiff_array_gradient(::Type{<:Complex},a,i_to_x,j_to_i::SkeletonPair;tag=default_tag(ForwardDiff.gradient,a))
```
JordiManyer, line 220: Same comment here, we can totally reuse a lot of things if we are clever. I would say we should split the skeleton work away from the merging of the blocks later, i.e the same thing we are doing [here](https://github.com/zjwegert/Gridap.jl/blob/cd97062787921e1a805257329e8fff642c7ffdb0/src/FESpaces/FEAutodiff.jl#L200). By creating this `_skeleton_autodiff_array_work` function, we can reuse everything without copying all the code again.

JordiManyer (changes requested):
Hi @zjwegert , thanks for the contribution. Very interesting, I didn't know you could do this. I added some comments on the reusability of the code. I think it would help to make the whole AD module more maintainable. Cheers!

zjwegert:
> Hi @zjwegert , thanks for the contribution. Very interesting, I didn't know you could do this. I added some comments on the reusability of the code. I think it would help to make the whole AD module more maintainable. Cheers!

Thanks @JordiManyer ! I'll take a look when I get a chance.

src/Arrays/Autodiff.jl
zjwegert: Agreed and added 👍

src/Arrays/Autodiff.jl
zjwegert: You're right. I've changed these.

src/FESpaces/FEAutodiff.jl
zjwegert: Agreed and added 👍

zjwegert:
Hey @JordiManyer, I've made the changes you requested and updated the NEWS. Cheers

zjwegert:
Hi @JordiManyer, this should be ready now. I'm happy to make changes in GridapDistributed once you're happy with this PR.

Edit: I've also added `adjoint` for ArrayBlocks

src/FESpaces/FEAutodiff.jl
```diff
@@ -42,11 +44,17 @@ end
 
 function _jacobian(f,uh,fuh::DomainContribution;tag::GridapADTag=get_ad_level(fuh)+1)
   terms = DomainContribution(;ad_level=tag)
+  V = get_fe_space(uh)
+  T = eltype(get_vector_type(V))
+  # For now, taking the Jacobian of a complex-valued functional is not implemented.
+  # To implement this, we need to consider the case of f being either holomorphic or
+  # non-holomorphic
+  T<:Complex && @notimplemented "Jacobian's of complex-valued functionals are not yet implemented"
```
JordiManyer: I think I would remove this check, and instead have placeholders 

```julia
function Arrays.autodiff_array_jacobian(::Type{<:Complex},....)
  @notimplemented
end

function Arrays.autodiff_array_hessian(::Type{<:Complex},....)
  @notimplemented
end
```

in the appropriate files.

src/FESpaces/FEAutodiff.jl
```diff
@@ -71,11 +79,14 @@ end
 
 function _hessian(f,uh,fuh::DomainContribution;tag::GridapADTag=get_ad_level(fuh)+1)
   terms = DomainContribution(;ad_level = tag+1) # Two levels consumed
+  V = get_fe_space(uh)
+  T = eltype(get_vector_type(V))
+  T<:Complex && @notimplemented "Hessian's of complex-valued functionals are not yet implemented"
```
JordiManyer: Same here

JordiManyer (changes requested):
Hi @zjwegert I have a couple style comments, but otherwise I think the changes are good. I will merge after!

PR #1326, opened 2026-07-13
Get ftype when change domain on empty adapt trian

ericneiva:
Hello @JordiManyer,

Here is a small PR with a fix when changing domain on empty adapted trians. The previous implementations of `change_domain`  where assuming the field was scalar. Now, it does not assume that and returns a `ZeroField`, instead of a `ConstantField`.

Lmk if you have any suggestions for improvement. Otherwise, thanks a lot for accepting the PR.

Cheers,
Eric


JordiManyer:
Looks good! Thanks @ericneiva for the contribution!

PR #1330, opened 2026-08-03
simplified BernsteinBasisOnSimplex

Antoinemarteau:
I removed the possibility of not storing a change of coordinate matrix (cartesian to barycentric) in BernsteinBasisOnSimplex.

This was an optimization for computations on reference simplices, but I realized it is complitely negligible and made the types of BernsteinBasisOnSimplex and BarycentricP(m)ΛBasis unecessarily complicated.

Now, the x_to_λ matrix is always stored, as a Matrix{Float64} instead of SMatrix, and SizedArray is used to cast the Matrix into a StaticArray to avoid allocations (and enable speed) in `_cart_to_bary`.

The Float64 hard-code leads to promotion to 64bits arithmetic for all evaluation of BernsteinBasisOnSimplex and P(m)Λ Bases. But the type parameters of both struct are way simpler now, the former B=BernsteinBasisOnSimplex{D,V,K} parameter is completely removed from P(m)Λ bases now.


Antoinemarteau:
@JordiManyer this is ready.
Both broken tests don't seam to be related to the changes as far as I can tell.

PR #1331, opened 2026-08-13
fix TensorValues elementwise operation error

wei3li:
Hi @JordiManyer, I believe this is a bug:
```cmd
julia> using Gridap

julia> a = TensorValue([1. 2. 3.; 3. 2. 1.])
TensorValue{2, 3, Float64, 6}(1.0, 3.0, 2.0, 2.0, 3.0, 1.0)

julia> a .* a
ERROR: AssertionError: The number of scalar arguments in TensorValue has to be a perfect square (e.g., 1, 4, 9, 16, ...)
```


Antoinemarteau:
Hey @wei3li, good catch thanks!

I used `@test_throws` instead of the try catch test, and took the opportunity to add broadcasting to `[Third/High]OrderTensorValue`.

@JordiManyer I think this is mergable if tests pass.

PR #1336, opened 2026-09-09
New moment-based reffes.

JordiManyer:
This PR adds many new ReferenceFE types, based on the moment and pullback machinery developed in #1048 . 

In particular: 

- Argyris 
- Morley
- HHJ, Regge and GLS
- Arnold-Winther (C/NC)
- Mardal-Tai-Winther

Moreover, I also added `CartProdReffes`. Given a base reference space `V` and an integer `k`, it represents `V^k = V x ... x V`, stacked column-wise (fast-index-wise). I also have a variant taking different spaces in different directions, that is `V1 x ... x VK` For instance, 

- `CP(Argyris, 3)` represents a vector-valued space with 3 components such that each component is Argyris.
- `CP(Raviart-Thomas{D}, D)` represents a (DxD)-tensor-valued space such that each column of the tensor is a D-dimensional RT space. 

Why did I decide to stack on the last index: If one looks in the literature, constructions such as `CP(Raviart-Thomas{D}, D)` usually rely on first-index stacking (that is each row would be RT). 
There are two reasons I decided to go with last-index stacking:

- First, because of how our gradients work. In Gridap, a gradient is `∇A[k,i,j] = ∂_k A_ij`. This means the derivative index is the first index. By stacking on the last index, the derivative and stacking operations commute, that is `∇ ∘ ι = ι ∘ ∇`. This would not happen if the stacked on the first index. 

- Second, Julia's last index is also the slow one. By last-index stacking, the components of each copy of `V` are contiguous in memory. Thus, it should be faster (and more elegant) to perform block-wise operations on each component.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `HHJRefFE`, `ReggeRefFE` and `GLSRefFE`: the Hellan-Herrmann-Johnson, Regge and Gopalakrishnan-Lederer-Schoberl (second kind) triangles, of any degree `r >= 0`, normal-normal, tangential-tangential and normal-tangential continuous respectively.
 - Added `DubinerBasis`, the L2-orthonormal (Dubiner/Koornwinder) polynomial basis of `P_K` on the reference simplex in any dimension.
 - Added `ArnoldWintherNCRefFE` and `ArnoldWintherCRefFE`, the nonconforming (15 DoFs, P2-based) and conforming (24 DoFs, P3-based) Arnold-Winther symmetric stress triangles for mixed elasticity. Both are built as constrained subspaces of a larger prebasis, so `length(get_prebasis(reffe)) > num_dofs(reffe)` for them.
+- Added `MardalTaiWintherRefFE`, the Mardal-Tai-Winther element on triangles (9 DoFs) and tetrahedra (24 DoFs).
 
 ## [Unreleased]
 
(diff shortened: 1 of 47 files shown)


JordiManyer:
@Antoinemarteau this is ready for a first review.

JordiManyer:
Are we good on this @Antoinemarteau ?

Antoinemarteau:
@JordiManyer Let me do a last review of the docs with minor improvements.

One thing on the implementation, I know feel like we should add those `[Curl/Div][Curl/Div]Conformity` and `:Hcurlcurl` etc., since we now have a curl-div element that is marked Hdiv only and I find it quite inelegant.. Do you mind if I add that ?

JordiManyer:
> @JordiManyer Let me do a last review of the docs with minor improvements.
> 
> One thing on the implementation, I know feel like we should add those `[Curl/Div][Curl/Div]Conformity` and `:Hcurlcurl` etc., since we now have a curl-div element that is marked Hdiv only and I find it quite inelegant.. Do you mind if I add that ?

Lets not, for now. In general I am having doubts about the whole `Conformity` thing. Other than L2 conformity, no other conformity really serves a purpose now other than "making the code pretty". They used to determine the pushforward, but we now just carry the pushforward around. So I would like to think a bit about this, instead of just blindly adding 100 new types that serve no purpose whatsoever.

Antoinemarteau:
@JordiManyer  I finished my review

Main changes:
- added Hermite element on simplices
- authorides `:L2` conforming reffes to go through the conforming fespace constructor when the cell does not own all the dofs, which is the case for AWnc (there is some face dof gluing even though the element is very non-conforming). It is cheap and small change in terms of LOC, but a "big" change in spirit. If this is bad, we should revert AWnc conformity to HDiv.
- added hessian evaluation of Dubiner bases
- added `HessConformity` alias `H2Conformity`

Minor changes:
- added `@tests` for convergence tests in tests/ConvergenceTests/ to avoid having to read outputs manually (no new resolutions)
- changed AWnc conformity to `:L2`, changed Argyris conformity to `:C1`.
- did many minor docs and code readability improvements

